PR #332, opened 2019-07-12
Adds REST API module to core

rwrobe:
Added a boilerplate for REST API modifications and registered a service provider to register the object with Pimple.


wp-content/plugins/core/src/REST_API/REST_API_Meta_Registration.php
```diff
@@ -0,0 +1,49 @@
...
+
+namespace Tribe\Project\REST_API;
+
+use Tribe\Project\Object_Meta\Example;
+use Tribe\Project\Post_Types\Sample\Sample;
+
+/**
+ * Class REST_API.
+ */
+class REST_API_Meta_Registration {
```
jbrinley: This is moving in the right direction, but I still contend that the class name is pushing us towards excessive responsibilities. I would prefer to see the name reflect exclusivity to the subset of meta that this class is concerned with. Should it be a single meta key per class? Or maybe a group of related meta keys? I'm not sure of the answer.

Taking a different approach, if we wanted an abstraction for the common case, we could have a class like: https://gist.github.com/jbrinley/19c8b2b267a2181396eb6a32f6321f69

Your service provider could then register instances of this class with different combinations of post types and meta keys. Any keys that needed special handling could extend this class to implement `get_value()` differently, OR we could make the constructor accept another arg for a `Value_Strategy` that we could pass in from the provider: https://gist.github.com/jbrinley/953cdf819ac3f203f2654526d0681021

PR #337, opened 2019-08-02
use mariadb instead of mysql

jbrinley:
Updates the global container to use MariaDB instead of MySQL.

After updating your local environment, a migration command is required:

```
docker exec -it tribe-mysql /bin/bash -c "mysql_upgrade -uroot -ppassword"
```

Otherwise MariaDB will throw some warnings at you (if you're looking in the logs).

### DIFF
--- a/dev/docker/global/docker-compose.yml
+++ b/dev/docker/global/docker-compose.yml
@@ -27,7 +27,7 @@ services:
         ipv4_address: 172.20.10.250
 
   mysql:
-    image: mysql:5.7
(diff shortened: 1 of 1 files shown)


jbrinley:
Per discussion, we're closing this. It doesn't offer any compelling gains, so we won't change what's not broken.

If anyone wants to run mariadb locally, they can set up their `docker-compose.override.yml` with:

```
version: '2.1'

services:

  mysql:
    image: mariadb:10

```

PR #338, opened 2019-08-02
set PAGER environment variable

jbrinley:
Sets the `PAGER` environment variable to `more` so that WP CLI commands that use the pager don't complain when `less` isn't available.

```
sh: 1: less: not found
```

Alternatively, we could just add `less` to our `Dockerfile`. I don't particularly have a preference between the two approaches.


jbrinley:
@speerface: could I get a quick review on this 1-line change?

PR #340, opened 2019-08-06
Feature/upgrade twig components

defunctl:
We're still using Twig classes that were deprecated in 2.7 and this updates our Twig classes to use the new namespaced classes.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -3,7 +3,7 @@
 All notable changes to this project will be documented in this file.
 
 ## 2019.08
-* Changed: sq1 Twig classes to use non-deprecated Twig classes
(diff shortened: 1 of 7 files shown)


defunctl:
> @defunctl: Does composer.json need to be updated to include the new version of Twig?

Shouldn't, nope. The lock file has v2.11.3 in it https://github.com/moderntribe/square-one/blob/master/composer.lock#L2565 these have been deprecated since v2.7

jbrinley:
`composer.json` says "^2.4". While the lock file has a newer version, `composer.json` indicates that our code will work with version 2.4. That should be updated to "^2.7" or "^2.11", if that's what we're coding for.

PR #341, opened 2019-08-06
R1 Button and CTA updates

jamiepaul:
Added button icon modifier. Refactored classes for button components to clean up modifier class application. Turned on Editor Formats dropdown for button variations.

Button icon modifier was applied as per Kyle's request here: https://tribe.slack.com/archives/CF7KVR8TT/p1564540878000900?thread_ts=1564511838.000700&cid=CF7KVR8TT


wp-content/themes/core/pcss/components/button/_outline.pcss
```diff
@@ -5,7 +5,8 @@
  * ----------------------------------------------------------------------------- */
 
 .c-btn-outline,
-a.c-btn-outline {
+a.c-btn-outline,
+a[data-mce-selected].c-btn-outline {
```
ryanurban: What is this `[data-mce-selected]` attribute?

wp-content/themes/core/pcss/components/button/_outline.pcss
```diff
@@ -14,29 +15,46 @@ a.c-btn-outline {
...
+	}
+
+	&[disabled] {
+		background-color: var(--color-outer-space);
+		border: 1px solid var(--color-nevada);
+		color: var(--color-nevada);
+	}
+
+	/* CASE: Button Modifier - Icon */
+	&.c-btn--icon:not([disabled]) {
```
ryanurban: Couple of questions:

* Do we need the additional `not` on this?
* Seeing this particular modified sprinkled throughout these button partials. Would it may make sense, given the primary purpose is to add an icon to a button to let it have it's own partial so all of it's styles could be together in one place? Let's further discuss this.

wp-content/themes/core/pcss/components/button/_solid.pcss
```diff
@@ -66,6 +109,7 @@ a.c-btn--inverse {
  * ------------------------------------------------------------------------- */
 
 .c-btn--full,
-a.c-btn--full {
+a.c-btn--full,
```
ryanurban: Do we use this anywhere in SQ1 at the moment? If not, I'd be inclined to remove it.

wp-content/themes/core/pcss/components/button/_outline.pcss
jamiepaul: @ryanurban Adding this shows the style immediately when a user selects a button style from the formats dropdown in a WYSIWYG editor. Example screenshots:
Without the attribute: http://p.tri.be/ApkUSE
With the attribute: http://p.tri.be/qBM7Vm

wp-content/themes/core/pcss/components/button/_solid.pcss
jamiepaul: Removed this.

wp-content/themes/core/pcss/components/button/_outline.pcss
ryanurban: Still not sure I follow on this. Let's do a quick call.

wp-content/themes/core/pcss/utilities/mixins/_buttons.pcss
```diff
@@ -20,93 +20,67 @@
...
- * Button - Solid
+ *
+ * Button - Repeated styles across all button components
+ *
  * ----------------------------------------------------------------------------- */
 
-@define-mixin button-solid {
-	background-color: var(--color-mine-shaft);
-	color: var(--color-white);
+@define-mixin button {
```
ryanurban: Similar as what we were doing before this iteration, I think it might make the most sense, for simplicity, to either not have mixin's for now for buttons or each type of button should get it's own mixin with it's core styles (for reuse) vs. having one global button mixin used across all buttons.

wp-content/themes/core/pcss/components/button/_outline.pcss
ryanurban: Thanks for the demo, that makes sense!

wp-content/themes/core/pcss/utilities/mixins/_buttons.pcss
ryanurban: Takeaway from call, rename this mixin to `button-shared`

PR #343, opened 2019-08-06
R1 WYSIWYG Panel

jamiepaul:
Ticket: https://central.tri.be/issues/131670
Design: https://moderntribe.invisionapp.com/d/main#/console/16185992/347968146/preview

Adds layout options to align panel header (left/center), with center being the default as this was previously the default in [1]. Implements design for all screen sizes.


wp-content/plugins/core/src/Panels/Types/Wysiwyg.php
```diff
@@ -12,6 +12,10 @@ class Wysiwyg extends Panel_Type_Config {
 	const FIELD_DESCRIPTION    = 'description';
 	const FIELD_COLUMNS        = 'columns';
 	const FIELD_COLUMN_CONTENT = 'column_content';
+	const FIELD_LAYOUT         = 'layout';
+
+	const OPTION_LAYOUT_LEFT   = 'option_wysiwyg_layout_left';
```
ryanurban: See here: https://github.com/moderntribe/square-one/pull/355#discussion_r313149269 and how that flows into the wrapper layout class ;)

wp-content/themes/core/pcss/content/panel/_wysiwyg.pcss
```diff
@@ -0,0 +1,39 @@
...
+		}
+	}
+
+	.g-col:not(:last-child) {
+		margin-bottom: var(--spacer-4);
+	}
+}
+
+.s-header {
+	@media (--viewport-medium) {
```
ryanurban: Remember that this is a globally available section utility class, so we should stay aware of whether the change we are making should be applicable only to this particular panel (sandbox it in your panel class) or is a global change (port to the sections header partial) and handle accordingly ;)

Fwiw, these do look like global changes.

wp-content/plugins/core/src/Panels/Types/Wysiwyg.php
jamiepaul: Updated

wp-content/themes/core/pcss/content/panel/_wysiwyg.pcss
jamiepaul: My mistake. I'm not sure the width should be global for this, so scoping to the panel until we see that pattern really emerge.

wp-content/plugins/core/src/Templates/Content/Panels/Wysiwyg.php
```diff
@@ -14,15 +20,64 @@ public function get_data(): array {
...
+
+	/**
+	 * Get WYSIWYG row wrapper classes.
+	 *
+	 * @return string
+	 */
+	protected function get_wrapper_classes() {
+		$classes = [];
+		$columns = $this->panel_vars[ Wysi::FIELD_COLUMNS ];
+
```
speerface: Let's check if $columns is empty here and return `Util::class_attribute( $classes, false )` if so. Then we can continue without having to check that in the next to conditionals.

speerface:
Let's make sure to add in a changelog entry for this too :)

wp-content/themes/core/content/panels/wysiwyg.twig
```diff
@@ -2,10 +2,26 @@
 
 {% extends "content/panels/panel.twig" %}
 
+{% block header %}
```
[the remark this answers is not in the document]
jamiepaul: Indeed. Good catch!

wp-content/plugins/core/src/Templates/Content/Panels/Wysiwyg.php
ryanurban: @speerface is that needed? Saw this on the project where it's getting implemented and seems like it's redundant given you still end up with the same return statement.

https://github.com/moderntribe/wstc/pull/1#discussion_r313949859

PR #352, opened 2019-08-09
🚝 Monorepo!

jbrinley:
Companion to https://github.com/moderntribe/tribe-libs/pull/29 to make Square One work with the new structure.

Reopens https://github.com/moderntribe/square-one/pull/339


wp-content/plugins/core/src/Core.php
```diff
@@ -100,23 +93,43 @@ private function load_service_providers() {
...
+		$optional_dependencies = [
+			'blog_copier'  => '\Tribe\Libs\Blog_Copier\Blog_Copier_Provider',
+			'container'    => '\Tribe\Libs\Container\Container_Provider',
+			'generators'   => '\Tribe\Libs\Generators\Generator_Provider',
+			'queues'       => '\Tribe\Libs\Queues\Queues_Provider',
+			'queues-mysql' => '\Tribe\Libs\Queues_Mysql\Mysql_Backend_Provider',
+		];
+
+		foreach ( $optional_dependencies as $key => $provider ) {
+			$this->providers[ $key ] = new $provider();
```
defunctl:
```suggestion
			if( class_exists( $provider ) ) { 
				$this->providers[ $key ] = new $provider();
			}
```

defunctl: I ran `./composer.sh remove moderntribe/square1-blog-copier` as a test and threw a 500. If these are optional, we should definitely just check if they exist before attempting to load them.

jbrinley: > If these are optional, we should definitely just check if they exist before attempting to load them

✅ done

PR #353, opened 2019-08-09
set PAGER environment variable

jbrinley:
Sets the `PAGER` environment variable to `more` so that WP CLI commands that use the pager don't complain when `less` isn't available.

```
sh: 1: less: not found
```

Alternatively, we could just add `less` to our `Dockerfile`. I don't particularly have a preference between the two approaches.

Re-opens https://github.com/moderntribe/square-one/pull/338


binarygary:
I prefer setting it as an env var like this instead of installing less. As you noted, there's no major benefit to either approach, but not installing a package is preferable to installing a package.

jbrinley:
I asked this before at entirely the wrong moment:

@speerface: could I get a quick review on this 1-line change?

PR #354, opened 2019-08-12
Add an .editorconfig for the core plugin

jbrinley:
This PR is meant to start a conversation. It's tangentially related to @rwrobe's work in https://github.com/moderntribe/square-one/pull/348.

Adding an `.editorconfig` file can help us share code formatting standards across the team. There aren't many universally supported properties, but PhpStorm has a large number of custom properties to set nearly ever aspect of code formatting for the project. This can allow us to configure team-wide settings on a per project or even per directory basis.

Some questions:

1. Do we want a single `.editorconfig` at the root of the project? Should we have one in the core plugin, another in tests, and another in the theme? Maybe there's a good way to have a single file but tell it to exclude certain directories (like 3rd-party plugins)?

2. There are a lot of rules here. There could be even more. We can add rules for more file types. We can add more PhpStorm properties. How detailed do we want to be? Does FE dev want in on this?

3. Is there a different system for sharing configs that we might find easier and preferable to this?

Y'all have other questions/comments/concerns?


rwrobe:
In favor of an `.editorconfig`. IMHO:

1. A single editorconfig seems best-- I won't be reformatting code in vendor files and can mark them as excluded if the warnings get annoying. I also like to see how far 3rd-party code is from my preferred codestyle, because the warnings can sometimes tell me a lot about code quality at a glance (assuming the code isn't obviously conforming to another major standard).  I also wonder if it's just worth standardizing codestyle across all our code, rather than excluding something like tests.

2. Would love it to be as complete as possible. Not sure if PHPStorm properties would cause problems in other IDEs or they'd just be ignored.

3. PHPStorm supports a settings repo, but it's always felt a little janky to me. Having changed workstations so often in the past, I just keep an export in Google Drive. Think `editorconfig` is the way to go and would love to see this!

PR #355, opened 2019-08-12
Feature/129674/media text boxed style panel

nicholasmelwell:
Create Media + Text Panel. Add support for Image or Video. Settings include Layout (Boxed or Full Bleed) and Media Position (Left or Right).

Awaiting icons from design.


wp-content/plugins/core/src/Panels/Types/MediaText.php
```diff
@@ -0,0 +1,93 @@
...
+	const FIELD_TITLE          = 'title';
+	const FIELD_DESCRIPTION    = 'description';
+	const FIELD_IMAGE          = 'image';
+	const FIELD_VIDEO          = 'video';
+	const FIELD_CTA            = 'cta';
+	const FIELD_MEDIA_TYPE	   = 'media_type';
+	const FIELD_MEDIA_POSITION = 'media_position';
+	const FIELD_LAYOUT         = 'layout';
+
+	const OPTION_MEDIA_TYPE_IMAGE   = 'option_media_type_image';
```
ryanurban: In seeing how this plays out, I like the namespacing of the strings less. We already get verbosity with the constant names. If we simplify these strings to be like: `right`, `left`, etc. then we also benefit from not having to do conditionals like this: https://github.com/moderntribe/square-one/pull/355/files#diff-39f42d7c872f2cdc85b67560c572c96eR256

wp-content/themes/core/content/panels/mediatext.twig
```diff
@@ -0,0 +1,27 @@
...
+{% block content %}
+
+    <div class="g-row g-row--no-gutters g-row--col-2--min-full {{ media_position }}">
+
+        <div class="g-col">
+            {{ media }}
+        </div>
+
+        <div class="g-col g-col--vertical-center">
+            <div class="t-content {{ media_spacing }}">
```
ryanurban: I believe `media_spacing` is now a relic before we changed course last week?

wp-content/plugins/core/src/Templates/Content/Panels/MediaText.php
```diff
@@ -0,0 +1,273 @@
...
+	 */
+	protected function get_classes(): string {
+		$no_padding = '';
+		$wrapper_class = $this->get_media_text_wrapper_class();
+
+		if ( MediaTextPanel::OPTION_LAYOUT_FULL_BLEED === $this->panel_vars[ MediaTextPanel::FIELD_LAYOUT ] ) {
+			$no_padding = 's-wrapper--no-padding';
+		}
+
+		$classes = [
```
ryanurban: Per our call last week, remember that we want to have parent wrapper modifier classes set around any options we have (image vs. video, left vs. right, full bleed vs. regular) for easy access.

wp-content/plugins/core/src/Panels/Types/MediaText.php
speerface: I concur. In ACF it's more necessary since those field names can conflict, but in the Panels context keeping the actual values generic should be just fine and dandy!

wp-content/plugins/core/src/Templates/Content/Panels/MediaText.php
```diff
@@ -0,0 +1,273 @@
...
+	/**
+	 * Overrides `get_classes()` from the Panel parent class.
+	 *
+	 * Return value is available in the twig template via the `classes` twig variable in the parent class.
+	 *
+	 * @return string
+	 */
+	protected function get_classes(): string {
+		$no_padding = '';
+		$wrapper_class = $this->get_media_text_wrapper_class();
```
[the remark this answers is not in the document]
speerface: I actually disagree :)

Any time I find myself doing additional logic inside a function, I find it's always best to break it out into another method so that each one can handle as few things as possible.

wp-content/plugins/core/src/Panels/Types/MediaText.php
nicholasmelwell: Sounds good!

wp-content/plugins/core/src/Templates/Content/Panels/MediaText.php
nicholasmelwell: Seems like that'll cause bloat, in my opinion. The `s-wrapper` has a modifier` s-wrapper--no-padding` which already removes padding from the wrapper to handle the full bleed and contained experience. why would we do this on a panel by panel basis? The image position is different since it directly affects the padding of the content container within the panel itself.

ryanurban: So last week we discussed how providing those "hooks" would be nice to have out of the box and give FE devs a variety of ways with which to quickly and efficiently target a given panel. I _think_ the pro of what it gives is greater than the con of a little bloat, but that's fair. Let's take a vote, good time to bring in the @frontend crew!

@frontend would it be of value to you to provide, out of the box, on the parent panel wrapper, a set of classes based on any options a panel has (thumb or down emoji)?

ckpicker:
@nicholasmelwell can you change the base of this PR to `feature/sq1-2019-theme-updates`?  🙇

PR #361, opened 2019-08-22
Feature/132625/gallery common panel

nicholasmelwell:
Create Dialog Component.
Create Gallery Controller and Gallery Panel Settings


wp-content/themes/core/pcss/components/_dialog.pcss
```diff
@@ -0,0 +1,87 @@
...
+	display: none;
+}
+
+.c-dialog__overlay {
+	background-color: var(--color-goethite);
+	bottom: 0;
+	color: var(--color-white);
+	left: 0;
+	overflow: scroll;
+	position: fixed;
```
ryanurban: For the sake of robusticity (new made up word) I prefer to have a wrapper inside of a parent fixed container that handles overflow scrolling sort of stuff. Gives us a greater degree of dependability on this fixed container may or may not interact in positioning when other site areas come into play.

wp-content/themes/core/pcss/components/_dialog.pcss
```diff
@@ -0,0 +1,87 @@
...
+	right: 0;
+	top: 0;
+	width: 100%;
+	z-index: 2;
```
ryanurban: Have we done any sort of z-indexery as vars, might be worth spinning something up so we can track our usage a bit (can probably wait for later, making a note).

wp-content/themes/core/pcss/components/_dialog.pcss
```diff
@@ -0,0 +1,87 @@
...
+	@media (--viewport-medium) {
+		padding-top: 140px;
+		width: 100%;
+	}
+}
+
+.c-dialog__content-inner {
+	margin: 0 auto;
+	padding: 0 20px;
+	max-width: 1000px;
```
ryanurban: Let's either add this to the other grid vars as an available max-width or pick one that exists that is close enough

wp-content/themes/core/pcss/components/_dialog.pcss
```diff
@@ -0,0 +1,87 @@
...
+		}
+	}
+}
+
+.c-dialog__header {
+	background-color: var(--color-goethite);
+	border-bottom: 1px solid var(--color-white);
+	margin-bottom: 47px;
+	padding: 20px;
+	position: fixed;
```
ryanurban: See above, going fixed on this in particular is fragile, depending on header title or whatever content may end up in the header from a height perspective. We can remove the fixed handling once you update the fixed item above and for now just allow the whole overlay to scroll, we'll do a refinement pass after this first pass on common panels). We could also potentially try to handle this as a flex situation where the header remains stuck and content fills up whatever area is left if that wouldn't take much time.

wp-content/themes/core/pcss/content/panel/_gallery.pcss
```diff
@@ -6,38 +6,110 @@
...
+		padding-bottom: var(--spacer-2);
+
+		@media (--viewport-medium) {
+			padding-bottom: var(--spacer-4);
+		}
+	}
+}
+
+.site-panel--gallery__content {
+	max-width: 848px;
```
ryanurban: Same notes on these max-widths or do you feel that these are so specific that they aren't worth moving into var form?

wp-content/themes/core/pcss/content/panel/_gallery.pcss
```diff
@@ -6,38 +6,110 @@
...
-		.tribe-image {
-			position: absolute;
-			width: 100%;
-			height: 100%;
-			top: 0;
-			left: 0;
-			background-position: 50% 50%;
-			background-size: cover;
+	.t-content & {
+		margin-top: 20px;
```
ryanurban: This should use an existing spacer var or choose the closest, we shall maintain spacing system integrity!

wp-content/themes/core/pcss/content/panel/_gallery.pcss
```diff
@@ -6,38 +6,110 @@
...
+	}
+}
+
+.site-panel--gallery__slider {
```
ryanurban: I _think_ we should turn a lot of the themery that is part of skinning this slider into a slider theme that can be used anywhere through the use of a c-slider--variation parent class. Let's chat.

wp-content/themes/core/pcss/utilities/variables/_colors.pcss
```diff
@@ -34,6 +34,7 @@
 	--color-nevada: #606a6b;
 	--color-submarine: #b9c2c3;
 	--color-desert-storm: #f2f1f0;
+	--color-goethite: #222222;
```
ryanurban: Micro optimizations ftw! `#222` 💨

wp-content/themes/core/pcss/content/panel/_gallery.pcss
nicholasmelwell: These are very specific to the panels that I've noticed so far. If we do do this, I'd recommend some consistency from design if possible. I'm seeing this mostly within the panel / section header areas with the panel title and description.

ryanurban: That's fair, could you follow up them and see, it's not a bad idea to have a consistent set in place ➕

speerface:
Same deal here - as soon as Ryan signs off, you can go ahead and merge, Nicholas!

ryanurban:
Closing this PR out for now as we'll need to update this following the greater reorg/rearchitecture of SQ1 and moving to blocks : )

Issue #362, opened 2019-08-22
TravisCI setup needs better env support


rwrobe:
TravisCI fails right now on builds because of issues with the tr1b0t oAuth token and the missing `.env` file in the Composer install step. We need to allow for env vars set in TravisCI to be available to Docker Compose.

defunctl:
I've begun to address a bunch of this in https://github.com/moderntribe/square-one/pull/351

nickpelton:
Moving to Github Actions for CI.

PR #351, opened 2019-08-09
Bugfix/travis tests

defunctl:
Adds WP_PLUGIN_ACF_KEY and WP_PLUGIN_GF_KEY encrypted environment variables and adds them the .env file so composer install can complete.

A bunch of tests are still failing and there are some other issues but the tests are running now: https://travis-ci.com/moderntribe/square-one/jobs/221493259#L1588

PR #363, opened 2019-08-22
Issue #362: TravisCI improvements

rwrobe:
I added some constants to the Docker Compose file's environment section and removed the tr1b0t oAuth token from travis.yml. You need to set this manually for the project in the TravisCI settings now, which is covered in the included documentation.

Please take note that the secret has been removed from the travisCI config-- this may cause unconfigured builds to begin to fail.


rwrobe:
Pleas see #351, which solves this issue by adding the secrets as an encrypted string, instead of relying on the env variables. This doesn't require any modifications to the docker yml, so I'm closing this PR in favor of #307

PR #369, opened 2019-09-30
Fix Tribe\Gutenpanels not being found

nealfennimore:
Was running into cannot find `\Tribe\Gutenpanels\Plugin` before regenerating the lock


nealfennimore:
Sounds good!

PR #371, opened 2019-10-09
Feature/improved search

Luc45:
While working on a new S1 project, I've stumbled upon a thing or two that I personally think would make sense to port to S1 as default. I'll just make the PR and let you decide that :rocket: :rocket: 

### Fix: Search button not working (https://github.com/moderntribe/square-one/commit/5369e872f2c8b7f988f1bae188233e30b9a98e28)
On a default S1 installation, clicking the search button does not trigger a search because "type=submit" is missing from the button attributes

### Improving search when no results found (https://github.com/moderntribe/square-one/commit/a88a2fc791d33f2fa4399624322bbd1f1890ac80)
By default on S1, the same template handles an empty archive of posts, (index.twig) and a search form where no results were found.

This commit aims to improve the defaults to a more sensible behavior, as seen on these screenshots:

| Before  | After with results | After without results |
| ------------- | ------------- | ------------- |
| image  | image  | image |

### DIFF
--- a/wp-content/plugins/core/src/Templates/Base.php
+++ b/wp-content/plugins/core/src/Templates/Base.php
@@ -155,7 +155,6 @@ class Base extends Twig_Template {
 	protected function submit_button(): string {
 
 		$btn_attr = [
-			'type'  => 'submit',
 			'name'  => 'submit',
 			'value' => __( 'Search', 'tribe' ),
 		];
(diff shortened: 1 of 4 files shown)


Luc45:
Perhaps we could add Pagination to search results by default.

What do you think @speerface, @ryanurban?

Luc45:
Closing this PR in favor of SquareOne 2.0.

After 2.0 has settled-in, I'll double-check to see if this is still relevant, and maybe open a new PR in the new structure.

PR #387, opened 2019-11-05
Feature/replace svg util

defunctl:
This PR removes our SVG_Support.php utility class and the Util Provider (as this was the only thing registered in it) and replaces SVG functionality with the Safe SVG plugin.

I've also updated WordPress to 5.2.4 and a number of plugins to their latest versions.


faction23:
Nice update, thanks for the upgrade on the svg game with security on uploads now.

PR #388, opened 2019-11-08
Bail if we're in panel preview

speerface:
This disables panels caching if we're in preview.


speerface:
Oh wow, A+ work here. I've seen some brilliant PRs in my time, but this one takes the cake. Well done, brave developer.

PR #390, opened 2019-11-21
Feature/fe build update

dpellenwood:
* Updates Node & NPM to latest LTS versions.
* Updates all FE build tooling to latest (compatible) package versions.
* Related misc FE build tooling tweaks to accommodate new package versions and address some bugs.


gulp_tasks/footer.js
```diff
@@ -5,7 +5,7 @@ const pkg = require( '../package.json' );
 module.exports = {
 	coreIconsVariables() {
 		return gulp.src( `${ pkg._core_theme_pcss_path }utilities/variables/_icons.pcss` )
-			.pipe( footer( '}' ) )
+			.pipe( footer( '}\n' ) )
```
dpellenwood: Just a few tweaks to the icon tasks here and in header, footer, and replace) to better format the copied css files.

gulp_tasks/postcss.js
```diff
@@ -10,6 +10,7 @@ const pkg = require( '../package.json' );
 const compilePlugins = [
 	require( 'postcss-partial-import' )( {
 		extension: '.pcss',
+		prefix: '_',
```
dpellenwood: the latest version of postcss-partial-import requires this value to find imports properly.

gulp_tasks/replace.js
```diff
@@ -7,16 +7,17 @@ module.exports = {
 		return gulp.src( [
 			`${ pkg._core_theme_pcss_path }base/_icons.pcss`,
 		] )
-			.pipe( replace( /url\('fonts\//g, 'url(\'var(--path-fonts)/icons-core/' ) )
+			.pipe( replace( /url\('fonts\/(.+)'\) /g, 'url(var(--path-fonts)/icons-core/$1) ' ) )
```
dpellenwood: This was necessary b/c the latest versions of postcss-custom-properties doesn't appear to be replacing properties within quoted strings, e.g. - `url('var(--path-fonts)/icons-core/')` doesn't get replaced with `url('/wp-content/themes/core/fonts/icons-core/')` when the task runs. Removing the quotes fixes it. And, per the CSS spec, quotes aren't required for `url()`'s.

gulp_tasks/replace.js
```diff
@@ -7,16 +7,17 @@ module.exports = {
...
 	},
 	coreIconsVariables() {
 		return gulp.src( [
 			`${ pkg._core_theme_pcss_path }utilities/variables/_icons.pcss`,
 		] )
 			.pipe( replace( /(\\[a-f0-9]+);/g, '"$1";' ) )
-			.pipe( replace( /\$icomoon-font-path: "fonts" !default;/g, '' ) )
-			.pipe( replace( /\$/g, '--' ) )
+			.pipe( replace( /\$icomoon-font-path: "fonts" !default;\n/g, '' ) )
+			.pipe( replace( /\$icomoon-font-family: "core-icons" !default;\n/g, '' ) )
```
dpellenwood: removed an additional sass var that doesn't get used.

gulp_tasks/shell.js
```diff
@@ -4,16 +4,16 @@ const browserSync = require( 'browser-sync' );
 
 module.exports = {
 	yarnInstall() {
-		return gulp.src( '' )
+		return gulp.src( './' )
```
dpellenwood: Heads up, the gulp-shell package has apparently been "blacklisted" by the gulp developers. And, `return gulp.src( '' )` is a big anti-pattern apparently. I tried a few attempts at using node exec/streams for these tasks instead of shell, but I wasn't successful. For now, this continues to work, but we should probably reevaluate the usage of gulp-shell in the future.

webpack/common.js
```diff
@@ -13,8 +13,6 @@ module.exports = {
 		],
 	},
 	module: {
-		noParse: /node_modules\/vex-js\/dist\/js\/vex.js/,
```
dpellenwood: Vex doesn't appear to be referenced in our codebase any longer and `exprContextCritical` is deprecated.

wp-content/themes/core/pcss/admin/editor-style.pcss
```diff
@@ -78,6 +58,26 @@ body {
 @import "../base/forms/controls/textarea";
 @import "../base/forms/validation/required";
 
+/* -----------------------------------------------------------------------------
```
dpellenwood: The latest versions of postcss-partial-import require imports to come before any other css declarations (like regular css requires). Hence, adjusting the order here.

PR #392, opened 2019-11-27
remove aws sdk from root composer because it conflicts with s3-upload…

nickpelton:
On the last three server setups I've done we use the `s3-uploads` plugin. It bundles an older version of the AWS SDK. We also have a more recent version of the AWS SDK in the core composer. I've removed it to prevent this conflict in the future. If someone needs a more recent version, they can add it per project and reconcile the version mismatch, rather than it always being a problem.


nickpelton:
Looks like it was added 2 years ago with a very descriptive comment "composer stuff" by Aaron B. https://github.com/moderntribe/square-one/blame/af8abad64cc72e4d9149c38a5e5423f2c7abedc4/composer.json#L63

nickpelton:
I see no usage within the core plugin.

nickpelton:
@speerface Can I merge this?

PR #393, opened 2019-12-04
Theme/related posts

marcovega:


wp-content/themes/core/pcss/base/grid/_columns.pcss
```diff
@@ -6,7 +6,6 @@
 
 .g-col {
 	min-width: 0;
-	position: relative;
```
ryanurban: Any particular reason for removing this? 😃

wp-content/themes/core/content/single/post.twig
```diff
@@ -1,32 +1,50 @@
...
 
-		</ul>
+			{{ include( 'components/share.twig' ) }}
 
-		{{ include( 'components/share.twig' ) }}
+		</footer>
+	</div>
 
-	</footer>
+	<div class="item-single__related-posts">
```
ryanurban: Couple of things here:

- Let's update this block of output to conditionally render based on whether or not there are actually related posts
- Let's update this wrapper to be a `section` vs div as it's more semantically appropriate

wp-content/themes/core/content/single/post.twig
```diff
@@ -1,32 +1,50 @@
...
+		<hr class="item-single__related-posts__divider">
+		<h2 class="item-single__related-posts__section-title">More Posts</h2>
```
ryanurban: We always make sure to namespace/localize any strings in our code so they are ready for translation out of the box: `{{ __( 'More Posts' )|esc_html }}`

wp-content/themes/core/content/single/post.twig
```diff
@@ -1,32 +1,50 @@
...
-			</li>
+			</div>
 
-			<li class="item-single__meta-author">
-				{{ __('by')|esc_html }}
-				<a href="{{ post.author.url|esc_url }}" rel="author">
-					{{ post.author.name|esc_html }}
-				</a>
-			</li>
+		</div>
```
ryanurban: Let's go ahead and wrap this markup in a `header` element 👍

wp-content/themes/core/content/single/post.twig
```diff
@@ -1,32 +1,50 @@
...
-	<footer class="item-single__footer">
-
-		<ul class="item-single__meta">
-
-			<li class="item-single__meta-date">
+	<div class="item-single__featured-image">{{ post.featured_image }}</div>
+
+	<div class="l-container__stagger item-single__content">
+		<h1 class="item-single__headline">{{ post.title }}</h1>
+		<div class="item-single__meta">
```
ryanurban: What do you think about updating this wrapper to be `<ul>` (and it's children accordingly) given it's a list of meta?

wp-content/themes/core/content/single/post.twig
```diff
@@ -1,32 +1,50 @@
...
+
+			{% if post.categories %}
+				<div class="item-single__meta__categories">
+					<ul class="item-single__meta__categories__list">
+						{% for category in post.categories %}
+							<li class="item-single__meta__categories__list-item">
+								<a class="item-single__meta__categories__category-link" href="{{ category.permalink }}">
```
ryanurban: Let's add a `rel="archives"` to these cat anchors ⚡️

wp-content/themes/core/content/single/post.twig
```diff
@@ -1,32 +1,50 @@
...
-
-			<li class="item-single__meta-date">
+	<div class="item-single__featured-image">{{ post.featured_image }}</div>
+
+	<div class="l-container__stagger item-single__content">
+		<h1 class="item-single__headline">{{ post.title }}</h1>
+		<div class="item-single__meta">
+
+			{% if post.categories %}
+				<div class="item-single__meta__categories">
```
ryanurban: Nice job here in terms of the adherence to BEM! One rec for you, to reduce the overall length that can wind up in our BEM class names, instead of the pattern of `parent__child__childofchild__ohnomorechildren`, which is also appropriate, we tend to do something like this: `item-single__meta-categories`, `item-single__meta-categories-list` 😄

wp-content/plugins/core/src/Theme/Image_Sizes.php
```diff
@@ -7,6 +7,8 @@
 class Image_Sizes {
 	const CORE_FULL      = 'core-full';
 	const CORE_MOBILE    = 'core-mobile';
+	const CORE_SQUARE    = 'core-square';
+	const CORE_LANDSCAPE = 'core-landscape';
```
ryanurban: Naming is always fun! We've been starting to follow a pattern around naming more based around aspect ratio these days (with the exception of core-full & core-mobile), so instead of: `CORE_SQUARE = 'core-square';` it would be `	const IMAGE_16X9 = 'image-16x9';`

This is a really handy tool I use all the time: https://andrew.hedges.name/experiments/aspect_ratio/

wp-content/plugins/core/src/Templates/Content/Single/Post.php
```diff
@@ -31,13 +40,18 @@ public function defer_get_content() {
...
-	protected function get_featured_image() {
+	protected function get_featured_image( $attrs = [] ) {
+		$default_attrs = [ 'src_size' => 'full', 'srcset_sizes' => [], 'srcset_sizes_attr' => null ];
+
 		$options = [
-			'wrapper_class' => 'item-single__image',
-			'echo'          => false,
+			'wrapper_class'     => 'item-single__image',
+			'echo'              => false,
+			'srcset_sizes_attr' => '(max-width: 767px) 300px, (min-width: 768px) 800px'
```
ryanurban: Fwiw, it's usually rare that you need to utilize this argument, unless you are doing some seriously art directed stuff 😄 Probably not necessary for now.

wp-content/plugins/core/src/Templates/Single.php
```diff
@@ -102,4 +106,42 @@ protected function get_pagination_item( $post_id ): string {
...
+		$cards = [];
+
+		foreach($posts as $post) {
+			$template = new Content\Single\Post( $this->template, $this->twig );
+			setup_postdata($post);
+			$data = $template->get_data( [
+				'featured_image' => [
+					'src_size'          => Image_Sizes::CORE_FULL,
+					'srcset_sizes'      => [ Image_Sizes::CORE_SQUARE, Image_Sizes::CORE_LANDSCAPE ],
+					'srcset_sizes_attr' => '(max-width: 767px) 300px, (min-width: 768px) 800px',
```
ryanurban: Same note

wp-content/plugins/core/src/Templates/Single.php
```diff
@@ -102,4 +106,42 @@ protected function get_pagination_item( $post_id ): string {
 		return $link->render();
 	}
 
+	protected function get_related_posts_cards() {
```
ryanurban: TBD on this and as we have discussed, it'll be part of a much larger convo and work. We'll also ultimately want BE to take a look and see what, if anything more, we'd like to do with related content as a thing as I could see this turning into a more globally available/reusable and robust thing 😃 

Nice job on getting in here and taking a first crack at it and also asking some great questions. Nice job in using the card component as well 🏅

wp-content/plugins/core/src/Theme/Image_Sizes.php
marcovega: Do you mean replace both for IMAGE_1X1 and IMAGE_16X9?

wp-content/plugins/core/src/Templates/Content/Single/Post.php
marcovega: Which one, echo?

ryanurban: 'srcset_sizes_attr'

wp-content/plugins/core/src/Theme/Image_Sizes.php
ryanurban: Yep

wp-content/themes/core/pcss/base/grid/_columns.pcss
marcovega: It would make child elements that are fixed or absolute to snap to the boundaries of g-col, because of our "categories" layout for related posts, they go all the way to the left in desktop, exceeding, but remain in position in mobile. Personally, I don't like position: relative anywhere unless we need it, against putting it somewhere as default and then do position: static. But was we don't have styles for related posts anymore, I'll revert this one.

ryanurban: That makes sense and I don't disagree. @nicholasmelwell do you happen to remember or see any issue with nixing this for this base grid helper class?

nicholasmelwell: Yep, no reason not to remove it. If it's necessary it can be added on a project-by-project basis.

wp-content/plugins/core/src/Templates/Single.php
```diff
@@ -102,4 +106,41 @@ protected function get_pagination_item( $post_id ): string {
...
+		global $post;
```
speerface: Hmm, is there a reason your calling global $post here and then overriding it in your foreach? Was WP doing the thing where certain post-related functions like get_the_content() weren't working without it?

ryanurban:
Closing this PR out for now as we'll need to update all of this following the greater reorg/rearchitecture of SQ1 : )

PR #395, opened 2019-12-09
Feature/postcss assets

dpellenwood:
* Replaces PostCSS custom property paths with a PostCSS plugin to resolve paths to the theme root.
* Updates PosCSS documentation and changelog
* Tweaks icons fonts replace task for better linter support (and fewer automated stylelint changes on dist build.)


package.json
```diff
@@ -156,6 +156,7 @@
     "mini-css-extract-plugin": "^0.8.0",
     "optimize-css-assets-webpack-plugin": "^5.0.3",
     "postcss-aspect-ratio": "^1.0.0",
+    "postcss-assets": "^5.0.0",
```
dpellenwood: @ryanurban I tried to use `postcss-url` first, but for the life of me could not get the options configured in a way that worked for our purposes. Probably spent ~3-4 hours playing with the configuration unsuccessfully before trying this as an alternative. 

Honestly, I like this plugin a little better because, as the developer, I have to make a conscience choice to use the plugin's `replace()` function instead of the traditional `url()` syntax. This mean no chance of accidental resolution/replacement/rebasing of a path that shouldn't be changed. (Just my 2-cents.)

dpellenwood:
> * Did you do a test run on how this plays with `postcss-inline-svg`?
Good thought. I did, yes. Since `@svg-load` is it's own function, it expects a `url()` argument. We can't change it to use `resolve()` instead. However, I don't think we need/want to. Since the SVG gets injected into the stylesheet, there's no need to "resolve" the differences in URLs between our different css base paths. Instead, you can always use a relative path (from the pcss file you're editing) for injecting an inline SVG, i.e. - `@svg-load icon-logo url('../../img/theme/icon-logo.svg') {}`

> * This more global setup is likely just fine, however I wonder if it might make sense to define a few more explicit paths (fonts, images, pcss in the theme are usually it) or if we might get any sort of micro performance boost in doing so? Again something that can likely wait until we play with this on our projects.
Good thought. I just ran an unscientific test between using the full theme path versus fonts, img, and pcss paths interdependently (`gulp postcss:theme` 3 times each way) and there was no noticeable difference between the two configurations. All six runs were in the ~1.8 seconds range. That being said, a really large theme (many files & directories) may benefit from adjusting. I agree we can watch this play out over time and adjust as needed.

One thing I like about the current config is that it's easy to remember that all `resolve()` definitions should start from the core theme directory as their root. If we configured the loadPaths to point at the img, fonts, and pcss directories inside the theme root, I have to remember to start my `resolve()` without that folder name. I.e - `resolve('fonts/icons-core/core-icons.woff2?xw14bx')` becomes `resolve('icons-core/core-icons.woff2?xw14bx')`.

package.json
ryanurban: Awesome, makes sense ⚡️

PR #396, opened 2019-12-16
Fix 'rel' attribtue.

ky13d3nt0n:
`rel` attribute is within `link` array.


ryanurban:
Thanks for the PR and bug fix @ky13d3nt0n 🎆

PR #401, opened 2020-01-31
Feature/tests improvements

Luc45:
**Add a bash script to run tests in the Container**
Perhaps the most useful feature of this PR is the two bash scripts to run tests inside the container, with or without Xdebug. It just makes it so easy, I've been using it a lot lately.

`./test.sh run acceptance` - Fast. Xdebug disabled.
`./testx.sh run acceptance` - Enables Xdebug on the context of WPLoader and any requests dispatched by WPBrowser or WPWebDriver.

More examples:

Running all suites with Xdebug enabled:
```
./testx.sh run acceptance
./testx.sh run webdriver
./testx.sh run integration
./testx.sh run unit
```

Other possibilities:
```
./test.sh run acceptance
./test.sh run webdriver
./test.sh run integration
./test.sh run unit

##
# Build the generated actors and tests classes. It already does that automatically
# whenever you run a suite, but you can also call it explicitly if you're developing
# on an actor or a tester class and want to see it autocompleted on some testcode
# before actually running the tests
##
./test.sh build

##
# Generate an integration test to test something
##
./test.sh generate:test integration Tribe/Project/Theme/Third_Party_Tags

##
# Generate a webdriver test to do something
##
./test.sh generate:test webdriver Can_Do_Something_Cest
```

See the full list of commands here: https://codeception.com/docs/reference/Commands

Notes:
- Due to some shenanigans necessary testx.sh to work on all suites, you can't use it for things other than running suites. For instance, if you'd like to run `generate` or `build`, you have to use `test.sh`.
- Running `test.sh` or `testx.sh` will trigger a `codecept clean`, which cleans the `tests/_data/_output` folder, which contains files generated during the last test run, such as screenshots or HTML snapshots (available in WebDriver and Acceptance suites, respectively)

**Use a dedicated Object Cache salt on tests-config.php**
I suspect that Object caching might be one of the main reasons why sometimes, tests, especially acceptance/webdriver, simply don't do what they're supposed to do even though everything looks right. This PR adds a dedicated cache salt on `tests-config.php` in order to isolate the object cache namespace for tests. I've added some notes about WPDb direct database manipulation and its effects on cache invalidation, so if the developer wants he can disable object caching completely on tests by following the instructions in the comments.

**Abstract modules into codeception.suite.yml**
Some suites can lead to repetitive and human-error prone configurations, especially WPLoader and their `activePlugins` list. So this PR abstracts the modules that repeat themselves throughout different suites into codeception.suite.yml, so other suites can just override any parameters that they'd like.

**Update dump.sql**
Updates dump.sql with the latest WordPress schema, so it doesn't ask for a "Database upgrade" when logging into the dashboard.

**Ignores _generated folder**
Files in _generated folder are generated by Codeception before each test run, it's like compiled CSS, there's no need to have it on the repo.

**Add Page Object examples**
Page Objects are a great way of abstracting HTML output from webdriver/acceptance tests. I've added a couple examples of webdriver tests using it. 

**Server name on PHP env**
Added `PHP_IDE_CONFIG=serverName=square1.tribe` to PHP env on Docker so PHPStorm can map the CLI request (as the ones made by WPLoader) to a local folder using Settings -> Servers path.

**Add a ValidateSuitesExtension**
I've written this extension to check for malformed suites, such as WPDb before WPLoader, or WPDb with WPLoader set to false, which are things that a dev can easily miss and have some headache trying to figure out.

**Other stuff**
There were other minor tweaks, details on "Files changed".

**Tests are passing**
Last but not least, fixes any failing test on a default S1 installation.


borkweb:
Quick question: How does this bash script differ from [the existing one](https://github.com/moderntribe/square-one/blob/master/dev/docker/codecept.sh)? Running tests in the container can be fired off with:

```bash
npm run codecept:test -- integration
npm run codecept:test -- acceptance
etc
```

_([via](https://github.com/moderntribe/square-one/blob/master/package.json#L47))_

borkweb:
Ah, the new bash scripts look to have some extra bits (one for running codecept with Xdebug on). Nice! If we are switching to theses new scripts, we should delete the old one and update/add the commands in package.json

dev/tests/tests/integration.suite.yml
```diff
@@ -2,32 +2,12 @@
...
-                - posts-to-posts/posts-to-posts.php
-            activatePlugins:
-                - core/core.php
-                - advanced-custom-fields-pro/acf.php
-                - posts-to-posts/posts-to-posts.php
-            bootstrapActions:
-                - [P2P_Storage, install]
-                - [P2P_Storage, init]
-                - tribe/project/queues/mysql/init_table
+            loadOnly: false
```
defunctl: Can we enable this by default? I've been using it in integration tests on my other projects and works well and much faster with it on.

tests-config-sample.php
```diff
@@ -1,132 +1,68 @@
...
+ * face unexpected results in tests that are hard to debug.
+ *
+ * If you'd like to disable object caching completely during tests:
+ * 1 - Replace the line below with "define( 'WP_CACHE_KEY_SALT', md5( microtime( true ) ) );"
+ * 2 - Call "if ( tribe_is_codeception() ) wp_suspend_cache_addition( true );" on a mu-plugin
+ *
+ * Generally speaking, it's a good idea to run tests with cache enabled
+ * because they run faster and you get to test your cache invalidation logic
+ */
+define( 'WP_CACHE_KEY_SALT', 'tests' );
```
defunctl: Have you fully tested this? I've run into so many problems without generating a uniqueID but perhaps that was an old bug.

tests-config-sample.php
```diff
@@ -1,132 +1,68 @@
...
 
-if ( empty( $table_prefix ) )
-	$table_prefix = 'tribe_';
+$GLOBALS['memcached_servers'] = [ [ 'memcached', 11211, ] ];
```
defunctl: We now use redis as the default object-cache.php, should add this in addition to the memcached global

```
/**
 * Redis
 */
$GLOBALS['redis_server'] = [
	'host'     => 'redis',
	'port'     => 6379,
	'auth'     => '',
	'database' => 0,
];
```

tests-config-sample.php
```diff
@@ -1,132 +1,68 @@
...
 
-// ==============================================================
-// Manually back up the WP_DEBUG_DISPLAY directive
-// ==============================================================
+/*
+ * Whoops
+ *
+ * If you enable Whoops, the Whoops error library will be used to provide better/prettier error logging.
+ */
+define( 'WHOOPS_ENABLE', true );
```
defunctl: Should this be enabled for tests?

wp-config.php
```diff
@@ -47,7 +47,23 @@ function tribe_getenv( $name, $default = null ) {
 // Conditionally load tests-config if the proper header or user agent is present
 // ==============================================================
 
-if ( ( isset( $_SERVER['HTTP_X_TRIBE_TESTING'] ) || ( isset( $_SERVER['HTTP_USER_AGENT'] ) && $_SERVER['HTTP_USER_AGENT'] === 'tribe-tester' ) || tribe_getenv( 'WPBROWSER_HOST_REQUEST' ) ) && file_exists( __DIR__ . '/tests-config.php' ) ) {
+/**
+ * Check if current request is being made by Codeception.
+ *
+ * @return bool
+ */
+function tribe_is_codeception(): bool {
```
defunctl: During the tests bootstrap process, would there be a way to set a define of something like `define( 'TRIBE_TESTS', true );` or some other flag that could be read here to avoid running these checks on every request?

defunctl:
Hey @Luc45,

Thanks for this!

Think you could get this working with Travis, and as Matt said the npm commands as well.

I had travis working in my other branch, but we'd need another secure token for the new `WP_PLUGIN_GF_TOKEN` environment variable: https://github.com/moderntribe/square-one/blob/2c1eb7a06e5e6acdc7fc7e8c5bfc3ec2e84b4517/.travis.yml

You should probably also enable automatic new line feeds in PhpStorm in Editor > General > `ensure line feed at file end on Save`

wp-config.php
Luc45: Good point! These checks were already there, I just refactored them into something more readable.

I did some benchmarks, though, and improved it a bit by delaying the `file_exists` check only on test requests.

1.  Running a function that just `returns true` 1 million times: 1 second
2.  Running `tribe_is_codeception` 1 million times, before refactoring: 6.4 seconds
3.  Running `tribe_is_codeception` 1 million times, after postponing `file_exists` only on test requests: 3.6 seconds
4.  Running `tribe_is_codeception` 1 million times,  after postponing `file_exists` only on test requests and using `getenv` instead of `tribe_getenv`: 2.6 seconds

If we could safely use `getenv` there instead of `tribe_getenv` we would get timings that are a little better. For a single request, though, any of those timings are negligible, even if `tribe_is_codeception` is called multiple times.

I've pushed a commit with scenario number 3.

tests-config-sample.php
Luc45: Good point! Set it to false by default

tests-config-sample.php
Luc45: I think you're right. Object cache makes sense in the context of end-to-end tests that replicate user behavior, such as webdriver. For those, we can run a dedicated object cache salt and test the cache validation logic.

However, when you throw into the mix tests that poke the internals to replicate certain scenarios, such as WPDb direct database manipulation and integration tests, which is not end-to-end, we might end up with an inconsistent object caching state. Perhaps we could disable object cache by default, and enable it for webdriver _if/when_ we'd like to test our caching invalidation logic.

I'd just point that, I'm under the impression that if we do use a unique cache salt for every request, we will end up using a lot of RAM, so we would probably want to call `if ( tribe_is_codeception() ) wp_suspend_cache_addition( true );` on a mu-plugin to prevent that

For now, I've added a unique object cache salt for each request, disabling object cache, in practical terms.

dev/tests/tests/integration.suite.yml
Luc45: Good point. I've added it as an explicit abstract parameter that we recommend each suite to define. If you'd like to explicitly set it to false by default, let me know and I'll change it again.

tests-config-sample.php
Luc45: It seems the default object cache on `master` is still Memcached. `tests-config-sample.php` basically mirrors `local-config-sample.php`, with only a few differences.

We could easily add Redis configuration to `tests-config-sample.php`, but since the default is still Memcached, it could be a bit confusing, don't you think?

Luc45:
> Ah, the new bash scripts look to have some extra bits (one for running codecept with Xdebug on). Nice! If we are switching to theses new scripts, we should delete the old one and update/add the commands in package.json

Great point Matt!

I've removed `codecept.sh` in favor of `test.sh` and `testx.sh`, and added them as npm commands:

- `npm run codecept:test -- webdriver`
- `npm run codecept:testx -- webdriver`

Runs webdriver tests with Xdebug on/off, respectively. Running tests without Xdebug can be much faster.

Indeed, the big difference from `test.sh` and `testx.sh`, despite giving you the ability of running tests faster, is that they enable you to control the Xdebug behavior, both in the context of the PHP CLI that the test themselves runs, but also from the requests dispatched by WPLoader and WPWebDriver, so you have Xdebug breakpoint on the code under test as well.

Luc45:
> Hey @Luc45,
> 
> Thanks for this!
> 
> Think you could get this working with Travis, and as Matt said the npm commands as well.
> 
> I had travis working in my other branch, but we'd need another secure token for the new `WP_PLUGIN_GF_TOKEN` environment variable: https://github.com/moderntribe/square-one/blob/2c1eb7a06e5e6acdc7fc7e8c5bfc3ec2e84b4517/.travis.yml
> 
> You should probably also enable automatic new line feeds in PhpStorm in Editor > General > `ensure line feed at file end on Save`

Thanks! I have enabled new line feed on save on PHPStorm.

I did a manual merge on your `bugfix/travis-tests` branch, but indeed, we'd need to re-generate the `WP_PLUGIN_GF_TOKEN` secure token and I have no idea how to do that. Can you help me with that one?

Also, please let me know if there's anything from the `bugfix/travis-tests` branch that you'd like me to port over, perhaps we can close that PR in favor of this one?

defunctl:
Tests are passing on Travis now.

Luc45:
wp-browser 2.2.36 was released last week and fixes what the `ValidateSuitesExtension` was fixing, so I suggest we remove it and bump the minimum wp-browser version from 2.2.14 to 2.2.36.

Here's a diff between the dependencies of 2.2.14 and 2.2.36: https://www.diffchecker.com/pmyW482H

Luc45:
Another point I would like to make is to lock the version of the Selenium Chrome image on Global, since running the `webdriver` suite with different Selenium versions might give erroneous results.

Eg., from: `image: selenium/standalone-chrome:latest`
To: `image: selenium/standalone-chrome:3.141.59`

This way it's easier for us to get consistent results, avoiding errors that are hard to debug such as:

- `[Facebook\WebDriver\Exception\UnknownCommandException] unknown command: Cannot call non W3C standard command while in W3C mode`
-  `[Facebook\WebDriver\Exception\InvalidArgumentException] invalid argument: missing command parameters`

These might have roots in different Selenium versions running the same tests.

I would love to hear your inputs on this, @speerface, @defunctl, @jbrinley

jbrinley:
> lock the version of the Selenium Chrome image on Global, since running the webdriver suite with different Selenium versions might give erroneous results

That's an excellent idea. The more we can do to make our test environment predictable and consistent, the better.

.travis.yml
```diff
@@ -91,8 +110,10 @@ install:
 before_script:
   - bash $TRAVIS_BUILD_DIR/dev/docker/exec.sh mkdir -p /application/www/wp-content/cache/twig
   - bash $TRAVIS_BUILD_DIR/dev/docker/exec.sh chown -R www-data:www-data /application/www/wp-content/cache
+  - bash $TRAVIS_BUILD_DIR/dev/docker/exec.sh chmod 777 /application/www/dev/tests/tests/_output
```
Luc45: @defunctl do we really need a 777 here? Could it be 755?

jbrinley:
This will merge into `master` as part of `epic/fidgety-feet`. Should we close this PR, or leave it open until then, @speerface?

speerface:
Let's close this one out; if nothing else it makes our # of PRs look artificially smaller.

jbrinley:
Closing, in anticipation of a future merge as part of `epic/fidgety-feet`.

PR #402, opened 2020-02-22
Feature/image component refactor

dpellenwood:
Refactors the image component per the following:
* Reduce complexity & differences between foreground (``) and background (`<div>`) implementations.
* Standardize const & key naming I.e. - always use `attrs` for multiple and `attr` for singular.
* Standardize the use of the `lazyload` class so it is applied consistently.
* Allow for usages where `<figcaption>` can be injected after the Image.
* Allow more consistent injection of classes & attributes as arrays like the components.
* Make use of our built-in `class_attribute()` and `array_to_attributes()` utility functions.
* Removed all usages of `the_tribe_image()` in the core plugin in favor of usingf the image component directly.
* Updated related docs for component and theme images.
* Updated changelog.md


wp-content/plugins/core/src/Templates/Content/Single/Post.php
```diff
@@ -31,13 +32,19 @@ public function defer_get_content() {
 		return apply_filters( 'the_content', get_the_content() );
 	}
 
-	protected function get_featured_image() {
+	protected function get_featured_image(): string {
+		$image_id = get_post_thumbnail_id();
+
+		if ( empty( $image_id ) ) {
```
ryanurban: Dumb q, does the image component not handle this on its own? Would it make sense for it to do so?

wp-content/themes/core/pcss/components/_image.pcss
```diff
@@ -5,6 +5,7 @@
  * ----------------------------------------------------------------------------- */
 
 .c-image {
+	position: relative;
```
ryanurban: What are you thinking this is for?

wp-content/plugins/core/src/Templates/Content/Single/Post.php
dpellenwood: Not a dumb question. I actually went back & forth on this myself.

The Image component (any of the components, actually) doesn't prevent you from rendering their markup without any data. I think this makes sense architecturally, and I didn't want to deviate from that construct in this instance.

Additionally, the image component can take an `IMG_URL` property instead of the `IMG_ID` and render a strait URL to an image file. Thus, we would have to check & bail only if both inputs were empty (minor change, of course).

wp-content/themes/core/pcss/components/_image.pcss
dpellenwood: we have a couple image component classes for different background image sizes (basically aspect ratio helpers) a little further down in this file. However, if the component (`c-image`) doesn't have position, the absolutely positioned background image will "bleed" to the next positioned container (by default the l-wrapper). Thus, I added a default position here so that the helper classes  can work out-of-the-box.  Happy to re-evaluate it. I don't know that I know of a solid system in place for how much is too much with default component styles.

ryanurban: Ah right, I was thinking that bgd image class is added to this same el. Carry on.

wp-content/plugins/core/src/Templates/Content/Single/Post.php
speerface: I also don't mind it being here - it allows us to bail before we do any processing at all within the Image component. I would be super-shocked if that was ever any sort of bottleneck, but I'm always a fan of not instantiating classes/processing if we don't need to :)

PR #403, opened 2020-02-22
Feature/color utils helper methods

dpellenwood:
Add a few helper methods to the theme Color class for use with ACF swatch fields and other such implementations.


wp-content/plugins/core/src/Theme/Resources/Colors.php
```diff
@@ -33,4 +33,44 @@ public static function get_color_options_subset( array $colors = [] ) {
...
 	}
 
+	/**
+	 * Create an array of color options for use with the ACF Swatch field.
+	 *
+	 * @param array $colors
+	 *
+	 * @return array
+	 */
+	public static function get_color_options_for_acf( array $colors = [] ) {
```
speerface: This is great! Is there any way we could change the method names so that they don't necessarily mention ACF? These methods seem like they could be useful for a whole host of things, so it'd be nice to generify them a bit.

wp-content/plugins/core/src/Theme/Resources/Colors.php
```diff
@@ -33,4 +33,44 @@ public static function get_color_options_subset( array $colors = [] ) {
...
+	 *
+	 * @param string $color
+	 *
+	 * @return string
+	 */
+	public static function get_color_key_from_acf_value( $color = '' ) {
+		$name = '';
+		foreach ( self::get_color_options() as $key => $value ) {
+			if ( in_array( $color, $value ) ) {
+				$name = $key;
```
speerface: You could also just `return` the $key here instead of tossing in a `break`

wp-content/plugins/core/src/Theme/Resources/Colors.php
dpellenwood: Good thought. I've tweaked the function names accordingly.

wp-content/plugins/core/src/Theme/Resources/Colors.php
dpellenwood: Good catch. Fixed.

PR #404, opened 2020-02-23
SQ1 + LoadBalancer != real user ip address

nickpelton:
We recently noticed Glomar's IP detection was not working. It was returning the Load Balancer or Proxy's IP address instead of the original users IP. While we could account for this in the Glomar itself, other plugins or code may not. This fix updates the REMOTE_ADDR server header value if we detect a HTTP_X_FORWARDED_FOR server header exists, which is what the LB and or Proxy is providing from the original users.

I assume this is safe as it's the recommended approach I've found. We could instead simply patch Glomar, but this is a silent issue and will likely happen again if we don't fix it at a higher level.

Debate welcome.

### DIFF
--- a/wp-config.php
+++ b/wp-config.php
@@ -70,6 +70,16 @@ if ( $_SERVER['HTTP_X_FORWARDED_PROTO'] === 'https' ) {
 	$_SERVER['SERVER_PORT'] = 443;
 }
 
+// ==============================================================
+// If a Load Balancer or Proxy is used, X-Forwarded-For HTTP Header to get the users real IP address
+// ==============================================================
+
+if ( isset( $_SERVER['HTTP_X_FORWARDED_FOR'] ) ) {
+	$http_x_headers = explode( ',', $_SERVER['HTTP_X_FORWARDED_FOR'] );
+
+	$_SERVER['REMOTE_ADDR'] = $http_x_headers[0];
+}
+
 $config_defaults = [
 
 	// Paths


wp-config.php
```diff
@@ -70,6 +70,16 @@ function tribe_getenv( $name, $default = null ) {
 	$_SERVER['SERVER_PORT'] = 443;
 }
 
+// ==============================================================
+// If a Load Balancer or Proxy is used, X-Forwarded-For HTTP Header to get the users real IP address
+// ==============================================================
+
+if ( isset( $_SERVER['HTTP_X_FORWARDED_FOR'] ) ) {
```
mitogh: Wonder if might be a good idea to integrate `http-Foundation` instead to do all the Request work as the library is pretty robust for example in this case `getClientIps()` is a good method to get the IP.

https://github.com/symfony/http-foundation/blob/0e96db44b3f84e6f3b02dc91d8f6798836974fc9/Request.php#L785-L794

- https://symfony.com/components/HttpFoundation

speerface: I would typically agree, but in this particular use-case the logic of extrapolating the IP address from the request itself is simple enough that I'm ok with doing it manually. 

I'd say the criteria for whether we stick to this are as follows:

- Is the HTTP_X_FORWARDED_FOR a reliable method for getting the correct IP?
- Is that header reliably structured so that exploding the value and nabbing the first entry in the resulting array works all the time? 
- Are there other areas where we're doing HTTP-based work that could also benefit from the foundation lib, thus making it more tenable for inclusion? 

At this point, I feel that the balance is tipped enough in favor of manual processing things that I'm comfortable with it. As the platform evolves, we may need to revisit it, but for now it seems like a sound approach.

mitogh: Agree I think at this point the library seems like too much for a small change :+1:

nickpelton: Agreed. Let's stick with this simple tweak for now.  If this proves to be unreliable in the future, we can def switch to a maintained lib.

PR #409, opened 2020-03-16
Bugfix/image component panel updates

jamiepaul:
Updates Image component constants within panel template controllers to conform to recently refactored Image component work. Essentially just a small update to ensure panels with background images are working out of the box.

Adds class `.c-image--bg` to be applied as a wrapper class for any background implementation needing the bg image to cover content positioned on top (i.e. Hero, Interstitials, etc).


dpellenwood:
Hey Jamie,

Thanks for this PR. I'm having a hard time wrapping my head around the best solution. I see how my refactor of the image component made some poor choices for background usage within the existing panels styles. My gut says that maybe we're throwing more code on to band-aid my bad code rather than a more comprehensive fix.

I _think_ it may be best to solve this by making some changes to both the image component and the default panel styles in concert. Before we do, it might make more sense to work out a list of the most common scenarios for the background image usage so we can address them all together. Maybe this is something we can address during our internal work in April.

For the time being, I'd suggest removing the `position: relative` I added to the `.c-image {}` style block. Then each of the default panel image declarations can be updated with fewer options & classes (like the following example from the hero panel controller) while still accomplishing the goal of this PR:

```
$options = [
	Image::IMG_ID          => $this->panel_vars[ HeroPanel::FIELD_IMAGE ],
	Image::AS_BG           => true,
	Image::USE_LAZYLOAD    => false,
	Image::ECHO            => false,
	Image::IMG_CLASSES     => [ 'c-image__bg' ],
];
```

Curious if you think the above suggestions work with your understanding of the panel usages.

nickpelton:
Hey @jamiepaul! Please merge or close the PR if it's no longer needed. Thanks!

jamiepaul:
Closing. There was a meeting between @dpellenwood, @ryanurban and myself followed by a slack discussion somewhere from BE that we should really solve this on a more global level. The conclusion was that this will not be merged and should just be closed out

PR #410, opened 2020-03-18
PHPCS Proposal

rwrobe:
This is a proposed PHPCS implementation/ruleset, including a GH action to run composer install and test.

Please be aware that the standard versions are not frozen by default; After creating a project, you would need to freeze these.

See #411 for a demo of what this does when you run PHPCBF.

### DIFF
--- a/.github/workflows/phpcs.yml
+++ b/.github/workflows/phpcs.yml
@@ -30,6 +30,7 @@ jobs:
         env:
           WP_PLUGIN_ACF_KEY: ${{ secrets.WP_PLUGIN_ACF_KEY }}
           WP_PLUGIN_GF_KEY: ${{ secrets.WP_PLUGIN_GF_KEY }}
+          WP_PLUGIN_GF_TOKEN: ${{ secrets.WP_PLUGIN_GF_TOKEN }}
         run: composer install --prefer-dist --no-progress --no-suggest
 
       - name: PHPCS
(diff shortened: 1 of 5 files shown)


rwrobe:
@JBCSU We need to add a GH oAuth key to the environment secrets with the key `GH_TOKEN`. Don't have perms to do it myself.

.github/workflows/phpcs.yml
```diff
@@ -0,0 +1,36 @@
...
+        uses: actions/cache@v1
+        with:
+          path: vendor
+          key: ${{ runner.OS }}-build-${{ hashFiles('**/composer.lock') }}
+
+      - name: Install Dependencies
+        if: steps.cache-dependencies.outputs.cache-hit != 'true'
+        env:
+          WP_PLUGIN_ACF_KEY: ${{ secrets.WP_PLUGIN_ACF_KEY }}
+          WP_PLUGIN_GF_KEY: ${{ secrets.WP_PLUGIN_GF_KEY }}
```
jbrinley: In addition to `WP_PLUGIN_GF_KEY`, we also need `WP_PLUGIN_GF_TOKEN` to use our GravityForms composer installer utility.

rwrobe: Good call, I added this in ea2473de!

rwrobe:
@jbrinley Composer versions have been updated. Not sure what happened there, but it should match master now, except for the PHPCS stuff in require-dev!

jbrinley:
Subsumed into #479 and merged

PR #415, opened 2020-03-26
Twig template refactor

jbrinley:
This is sort of work in progress, in the sense that there is still refactoring to do, especially around documentation, DI containers, and the `Core` class. More will come in the next day or so in another PR, but we want to get this merged ASAP to avoid horrible conflicts down the line.

@speerface, can you take a slightly-more-than-cursory look at this, even if it's already been merged?


wp-content/plugins/core/src/Templates/Controllers/Comments.php
```diff
@@ -1,11 +1,19 @@
...
+use Tribe\Project\Templates\Abstract_Template;
+use Tribe\Project\Templates\Component_Factory;
+use Twig\Environment;
 
-use Tribe\Project\Twig\Twig_Template;
+class Comments extends Abstract_Template {
+	protected $path = 'comments.twig';
+
+	public function __construct( Environment $twig, Component_Factory $factory ) {
+		parent::__construct( $twig, $factory );
```
speerface: Any particular reason we're leaving the constructor override in-place with only the parent call in here?

jbrinley: Nope!

wp-content/plugins/core/src/Templates/Controllers/Content/Panels/CardGrid.php
```diff
@@ -82,18 +82,23 @@ protected function get_cardgrid_attributes() {
 
 	protected function get_card_image( $image_id ) {
 		if ( empty( $image_id ) ) {
-			return false;
+			return '';
+		}
+
+		try {
```
speerface: I like this pattern; do we think there's any value in also logging the Exception for troubleshooting purposes? Otherwise it just sorta fails silently.

wp-content/plugins/core/src/Templates/Controllers/Page.php
```diff
@@ -0,0 +1,166 @@
...
+	/**
+	 * @var Main_Sidebar
+	 */
+	private $sidebar;
+	/**
+	 * @var Footer
+	 */
+	private $footer;
+
+	public function __construct(
```
speerface: I _really_ like this change - injecting the other controllers as dependencies is a substantial improvement, and with our DI autowiring it's super-clean. :standing-ovation:

wp-content/themes/core/base.twig
```diff
@@ -1,15 +1,14 @@
...
 {% endblock %}
 
 <a id="a11y-skip-link-content" tabindex="-1"></a>
 
 <main>
-
 	{% block main %}
 
 		{% block subhead %}
-			{{ include( 'content/header/sub.twig' ) }}
```
speerface: Ooooh is the goal here to move away from using twig `include()` and always just pass sub-templates from the Controller? I like that.

PR #420, opened 2020-04-06
Allow us to set `opcache.validate_timestamps=0` in php.ini

defunctl:
Just did this for LL in https://github.com/moderntribe/little-league/pull/428

I think it should be the standard config in the upcoming Template refactor.

### DIFF
--- a/wp-content/plugins/core/src/Twig/Twig_Definer.php
+++ b/wp-content/plugins/core/src/Twig/Twig_Definer.php
@@ -6,6 +6,7 @@ namespace Tribe\Project\Twig;
 use DI;
 use Psr\Container\ContainerInterface;
 use Tribe\Libs\Container\Definer_Interface;
+use Twig\Cache\FilesystemCache;
 use Twig\Environment;
 use Twig\Loader\FilesystemLoader;
 use Twig\Loader\LoaderInterface;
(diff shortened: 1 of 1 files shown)


jbrinley:
I _think_ this makes sense, but can you provide a little more context about what this is accomplishing, and why it should be the default?

defunctl:
> I _think_ this makes sense, but can you provide a little more context about what this is accomplishing, and why it should be the default?

Absolutely. On production environments (where we have this control), we should really be disabling opcache timestamp validation: https://www.php.net/manual/en/opcache.configuration.php#ini.opcache.validate-timestamps.

While someone may remember to configure this in php.ini, it would cause all kinds of unintended caching with Twig, just enabling this option could save a number of headaches down the road.

Given Tribe Cache also calls `opcache_reset()` this would properly flush the Twig cache with no timestamp validation enabled.

More reading: https://twig.symfony.com/doc/2.x/recipes.html#refreshing-modified-templates-when-opcache-or-apc-is-enabled

jbrinley:
Are there any negative side effects if we have this value set on a server that does _not_ use `opcache.validate_timestamps=0`?

defunctl:
> Are there any negative side effects if we have this value set on a server that does _not_ use `opcache.validate_timestamps=0`?

Very good question, and I don't now that answer 100%. It basically asks Twig to invalidate the opache when writing a file with error suppression: https://github.com/twigphp/Twig/blob/3.x/src/Cache/FilesystemCache.php#L64

jbrinley:
Seems safe. 🤷‍♂

defunctl:
Let me know if you run into any issues.

PR #422, opened 2020-04-06
Server/dev

nickpelton:
Pipeline deploy for the core SquareOne Repo. This pipeline is tied to each server/ branch respectively and supports a dev/staging/production environment. It's running Jenkins for Build and Deploy orchestration and using a combination of Jenkins and Ansible vault for environment secrets.  For example, any code commit to a server branch will trigger the respective environment build. It assumes it's being deployed with a GIT Deploy pattern like common WordPross hosts like WPE and Pantheon provide. In the future I'd like to move primarily to GH actions, but this is a good implementation for now.

Other stuff:
- It also bundles an SH script for local if Jenkins were to fail. This is a handy backup.
- It also has a core plugin update that supports the autoloader different install configurations.


jbrinley:
Does this allow any possibility to deploy arbitrary branches, especially to dev or staging? That's a pretty key requirement for most projects.

nickpelton:
> Does this allow any possibility to deploy arbitrary branches, especially to dev or staging? That's a pretty key requirement for most projects.

@jbrinley It's currently pipeline only, which would contribute well to shifting to a CI/CD flow. But, it is possible to add custom deploys by adding some parameters. I propose we only allow custom deploys to dev while staging and prod are pipeline only. Thoughts?

jbrinley:
> I propose we only allow custom deploys to dev while staging and prod are pipeline only

That seems reasonable for most cases. Any exceptions can be handled per project.

Jenkinsfile
```diff
@@ -0,0 +1,178 @@
...
+                slackSend(channel: "${SLACK_CHANNEL}", message: "Pipeline: (<${RUN_DISPLAY_URL}|#${BUILD_NUMBER}>)\nDeployment of `${APP_NAME}` branch `${env.BRANCH_NAME}` to `${env.ENVIRONMENT}` STARTED.")
+               // checkout scm
+                checkout([$class: 'GitSCM',
+                    branches: [[name: "${env.BRANCH_NAME}" ]],
+                    extensions: [[$class: 'WipeWorkspace'], [$class: 'RelativeTargetDirectory',  relativeTargetDir: BUILD_FOLDER]],
+                    userRemoteConfigs: [[url: "git@github.com:${env.GIT_REPO}", credentialsId: "jenkins-ssh-key"]]
+                ])
+            }
+        }
+        stage('Deploy SCM') {
```
thejaywhy, line 30: this stage isn't really doing a `Deploy`, right? Maybe `Checkout Remote SCM` or similar?

Jenkinsfile
```diff
@@ -0,0 +1,178 @@
...
+                withCredentials([string(credentialsId: "${JENKINS_VAULTPASS}", variable: 'vaultPass')]) {
+                    sh script: "echo '${vaultPass}' > ./.vaultpass", label: "Write vaultpass to local folder"
+                    sh script: "ansible-vault decrypt ${env.BUILD_FOLDER}/dev/deploy/.host/config/${env.ENVIRONMENT}.cfg.vaulted --output=${env.BUILD_FOLDER}/dev/deploy/.host/config/${env.ENVIRONMENT}.cfg --vault-password-file ./.vaultpass", label: "Decrypt config config file"
+                    sh 'rm ./.vaultpass'
+                }
+
+                // Load Host environment variables
+                loadEnvironmentVariables("${env.BUILD_FOLDER}/dev/deploy/.host/config/${env.ENVIRONMENT}.cfg")
+
+                // checkout scm WPEngine
```
thejaywhy, line 42: Might want to change this to `WPEngine, Pantheon, or similar git-based deployment`

dev/deploy/deploy_hosted_git.sh
```diff
@@ -0,0 +1,183 @@
...
+
+function console_log(){
+  echo ""
+  echo "#########################################################################"
+  echo "### $1"
+  echo ""
+}
+
+if [ ! -f "./deploy_hosted_git.sh" ]; then
+    echo "Script must be run in the /dev/deploy/ directory. Aborting..."
```
thejaywhy, line 11: the warning message doesn't describe what the check is actually checking for.

PR #429, opened 2020-04-07
add a class to filter the block blacklist

jbrinley:
@faction23: per our brief discussion in Slack, this has a JS component, and I'm not sure where to put it.

1. On the PHP side, we're generating a list of block types that should be blacklisted. That list should be passed to the front end, via `wp_localize_script()`. I'm not sure if I should add it on to our existing [admin js config](https://github.com/moderntribe/square-one/blob/epic/fidgety-feet/wp-content/plugins/core/src/Admin/Resources/JS_Config.php#L15) or if it should be attached to a different script.

2. Once the block editor initializes, we need to call `wp.block.unregisterBlockType()` for each of the block types in the list. See https://developer.wordpress.org/block-editor/developers/filters/block-filters/#using-a-blacklist

Something along the lines of:

```
import { unregisterBlockType } from '@wordpress/blocks';

BLOCK_BLACKLIST.map( type => unregisterBlockType(type) );
```


faction23:
Yes please lets add it to the admin js config. Tomorrow i will map wordpress admin js in the admin webpack bundle as externals and wire that up. We just have some other work in related areas going on we need to resolve first!

PR #430, opened 2020-04-08
Block editor styles

jbrinley:
Enqueues the full front-end stylesheet for the block editor, keeping the existing `editor-style.css` for the classic editor.

### DIFF
--- a/wp-content/plugins/core/src/Theme/Resources/Editor_Styles.php
+++ b/wp-content/plugins/core/src/Theme/Resources/Editor_Styles.php
@@ -25,7 +25,7 @@ class Editor_Styles {
 	 *
 	 * @filter tiny_mce_before_init
 	 */
-	public function visual_editor_body_class( $settings ) {
+	public function mce_editor_body_class( $settings ) {
 
 		$settings['body_class'] = ( $settings['body_class'] ?? '' ) . ' t-content';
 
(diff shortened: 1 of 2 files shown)


wp-content/plugins/core/src/Theme/Resources/Editor_Styles.php
```diff
@@ -33,4 +32,45 @@ public function visual_editor_body_class( $settings ) {
...
+	 */
+	private function theme_stylesheet_path(): string {
+		return 'assets/css/dist/theme/master' . $this->file_suffix();
+	}
+
+	/**
+	 * @return string The relative path to the MCE stylesheet
+	 */
+	private function mce_stylesheet_path(): string {
+		return 'assets/css/dist/admin/editor-style' . $this->file_suffix();
```
faction23, line 67: I still feel like i should rename this to tinymce-style as thats what it applies to, i can do it after merge in next work, cool? Then we can also break up more with the block styles etc as we go. "Editor style" is now confusing.

wp-content/plugins/core/src/Theme/Resources/Editor_Styles.php
```diff
@@ -6,24 +6,23 @@
...
+		add_theme_support( 'editor-styles' );
 	}
+
 	/**
-	 * Visual Editor Body Class
+	 * Add a body class to the visual editor
+	 *
 	 * @filter tiny_mce_before_init
 	 */
 	public function visual_editor_body_class( $settings ) {
```
faction23, line 28: Is 'visual_editor' still the right name for these? mce_editor instead maybe? Let me know whatever you think is right, and then i will sync the css.

jbrinley: I think `mce-editor` is the right name. I didn't want to change it here until you changed the stylesheet file name. But I guess it doesn't really matter; I'll go ahead with the change for the function names, at least.

faction23: ok! I will modify that when i do your block blacklist work, which i could get to tonight actually after you merge.

PR #434, opened 2020-04-10
Set colors, gradients, font sizes for block settings

jbrinley:
The block editor supports overrides of all of the colors, gradients, and font sizes presented to users in block settings.

We will define all of these in the `Theme_Definer`. We are, by default, disabling the ability for users to set arbitrary values for any of these. Returning an empty array for any of these will disable it completely.

To discuss: how do we want to pass these values from PHP to the theme? Previously, we have duplicated the values between the core plugin and a CSS a file in the theme. Is that still optimal, or is there a better solution we could use to define the options only once?


wp-content/plugins/core/src/Theme/Config/Font_Sizes.php
```diff
@@ -0,0 +1,65 @@
...
+
+	/**
+	 * Formats the sizes as `[ 'name' => <label>, 'slug' => <key>, 'size' => <size> ]`
+	 *
+	 * This is the format expected by the block editor.
+	 *
+	 * @return array
+	 * @see https://developer.wordpress.org/block-editor/developers/themes/theme-support/#block-font-sizes
+	 */
+	public function format_for_blocks(): array {
```
dpellenwood, line 53: I may not understand completely, but I don't think we want to use an actual font-size control since our sizes are generally different per viewport size; and, we apply sizes based on css classes (or specific html elements inside a css class).

wp-content/plugins/core/src/Theme/Theme_Definer.php
```diff
@@ -6,17 +6,102 @@
 use DI;
```
dpellenwood, line 6: Love the consolidation of all theme config into a single class. Can we see about moving webfont definitions here as well?

jbrinley: Yes, fonts are on my future roadmap. They don't apply so specifically to the block editor, so I was tackling these things first.

wp-content/plugins/core/src/Theme/Config/Font_Sizes.php
jbrinley: These are the sizes that a user can select in the block editor for, e.g., paragraph blocks. It functions by applying a class to the paragraph, so the theme could handle it differently depending on viewport, etc. The font sizes we give are for admin display purposes only.

These are entirely optional. If we don't want to give users the ability to shrink/embiggen text, we can just return an empty array here.

PR #436, opened 2020-04-10
Feeture/dead components

dpellenwood:
Removes some duplicate (Title) or unused (Wysiwyg) components.


dpellenwood:
> Looks good. I'm sure you think about this as you go, but I'd love for us to standardize the arg naming for our components. Only question I have is on the WYSIWYG component, Davee, I recall you and I talking about having this and having it be the component that implements true WYSIWYG (would include `t-content` by default, etc.)?

Yeep. I'm not sure we'll call it the same thing though. For now, I want to eliminate the old one. We can evaluate the need for something along those lines as we get into building the new components.

PR #439, opened 2020-04-13
Comment templates

jbrinley:
Refactors the single comment template to use twig instead of the `core_comment()` template tag.


wp-content/themes/core/components/comments/comment/Comment.php
```diff
@@ -0,0 +1,87 @@
...
+		],
+	];
+
+	public function get_data(): array {
+		$this->properties[ self::ATTRIBUTES ][ self::MERGE_ATTRIBUTES ]['id'] = sprintf( 'comment-%d', $this->comment_id );
+
+		$data = parent::get_data();
+
+		// massage the timestamp into the formats we need in the template
+		$data['time'] = [
```
dpellenwood, line 77: Question I've had: any chance we can avoid having to define the format(s) we want in the context and use a helper function in the twig templates to render a specific format instead? A kind of pass-through from php's `date()`?

jbrinley: The thought has crossed my mind, too, @dpellenwood. We could consistently pass in a timestamp and render whatever format from the template...

Yes, let's do that. Different PR to come for that.

PR #441, opened 2020-04-13
Remove theme template tags

jbrinley:
Eliminates the two remaining template tags in the theme.

- `get_page_title()` - only used in one place, the `Head` controller. This code has been moved to that class.

- `the_tribe_image()` - only used in obsolete documentation. I've updated the docs for now, but they'll be deleted eventually, anyway.


wp-content/themes/core/functions.php
```diff
@@ -1,29 +1,6 @@
...
- */
-
-
-// Core Theme Plugin
-add_action( 'after_setup_theme', 'core_theme_setup' );
 
 /**
- * Theme Core Functionality "Plugin"
+ * Nothing should be here. If you are tempted to add functions here, consult with your
+ * team to determine how best to include that functionality in the core plugin.
```
dpellenwood, line 5: :tada:

PR #442, opened 2020-04-13
New! Link Component

dpellenwood:
* Adds a new Link component.
* Updates all BUTTON_AS_LINK usages to use the new Link component.
* Updates properties & attributes for Button & Link to new syntax.
* Updates component templates to drop the hard-coded `class={{...}}` attribute markup in favor of using the classes Util method.
* Fixes a few mixed multiple strings inside array class usages.


wp-content/themes/core/components/link/link.twig
```diff
@@ -0,0 +1,5 @@
+<a {{ classes }} {{ attrs }}>
+	{{ prepend }}
+	{{ text }}
+	{{ append }}
```
jbrinley, line 4: What is the purpose of separate `prepend`, `text`, and `append` variables? This has the same effect as including everything in `text`.

dpellenwood: Fair question. It's to provide a separation for times when we do something like:
`<i class="icon icon-phone"></i><span class="...">987-321-6541</span>`, where the prepended markup for the icon should be separated from the hover-able/focus-able "text" label. I realize this would be possible by concatenating the icon markup with the text label markup inside a single variable; but it provides additional flexibility this way to not have to contact those further "up" the chain.

jbrinley: > the prepended markup for the icon should be separated from the hover-able/focus-able "text" label

That's not what's happening here, though, is it? The icon and the text are all squashed together inside the anchor.

dpellenwood: Yes, that's precisely the point. In my example, we want the icon to be part of the clickable link, but not included in the hover/focus styles (underlining, for example).  The whole link (or button) component markup would be:
`<a href="..."><i class="icon icon-phone"></i><span class="...">987-321-6541</span></a>`

jbrinley: I still don't get it. How would you use this component to accomplish this?

```
$this->factory->get( Link::class, [
  Link::URL     => $url,
  Link::PREPEND => '<i class="icon icon-phone"></i><span class="...">',
  Link::TEXT    => '987-321-6541,
  Link::APPEND  => '</span>',
] )->render();
```

How is that better than this?

```
$this->factory->get( Link::class, [
  Link::URL  => $url,
  Link::TEXT => sprintf( '<i class="icon icon-phone"></i><span class="...">%s</span>', '987-321-6541 ),
] )->render();
```

jbrinley: If the template were something more like this, I could understand some of the logic behind it.

```
<a {{ classes }} {{ attrs }}>
	{{ prepend }}
	<span {{ inner_classes }}>{{ text }}</span>
	{{ append }}
</a>
```

dpellenwood: The idea is to separate the prepended (or appended) icon markup from the link text markup:
```
$this->factory->get( Link::class, [
	Link::URL     => $url,
	Link::PREPEND => '<i class="icon icon-phone"></i>',
	Link::TEXT    => sprintf( '<span class="...">%s</span>', '987-321-6541' ),
] )->render();
```

dpellenwood: Similar to how we specifically call out `{{ classes }}` from `{{ attrs }}` in the templates, there are specific cases where this will come into play, and it aligns with how we want to be able to name the component's css: https://9elements.com/bem-cheat-sheet/#button

jbrinley: You can still do that with just the `text` property. What I see here is a need for an additional component that you will sometimes inject into a `Link` as its `text`.

```
$simulated_button = $this->factory->get( Simulated_Button::class, [
  Simulated_Button::PREPEND => '🚀',
  Simulated_Button::BODY    => __( 'My Button', 'tribe' ),
  Simulated_Button::APPEND  => '👨‍🚀',
] )->render();

$link = $this->factory->get( Link::class, [
  Link::URL     => $url,
  Link::TEXT    => $simulated_button,
  Link::CLASSES => 'button',
] )->render();
```

Where `simulated-button.twig` looks something like:

```
  <span class="button__prepend">
     {{ prepend }}
  </span>
  <span class="button__body">{{ body }}</span>
  <span class="button__append">
     {{ append }}
  </span>
```

Otherwise, we're producing a lot of markup back in the controller again.

dpellenwood: Sorry, I wasn't intending for those to be complete examples. The markup in the example could be generated by any number of other components, yes. A "simulated button" is overkill and not realistic from a FE perspective, imho.

The markup for prepend, text, & append may not always follow this _exact_ pattern. FE needs the flexibility to render different markup, classes, and "text" within any given Link or button component. if we end up going the route of making a "simulated button" style for every possible FE link or button implementation, there would be dozens if not hundreds of "simulated buttons" on any given project.

jbrinley: My point is that you already have that flexibility. Removing those arguments from the component would not change it in the least. There is functionally no difference between these:

```
sprintf( '%s%s%s', $prepend, $text, $append );
```

```
$prepend . $text . $append
```

dpellenwood: @jbrinley let me know if I'm good to merge here when you have time. :)

PR #443, opened 2020-04-13
Fix accordian key and value for image options

nealfennimore:


wp-content/plugins/core/src/Blocks/Types/Accordion.php
```diff
@@ -37,8 +37,8 @@ private function layout_sidebar(): Sidebar_Section {
 			->add_field(
 				$this->factory->sidebar()->field()->image_select( self::LAYOUT )
 					->set_label( __( 'Layout', 'tribe' ) )
-					->add_option( self::LAYOUT_STACKED, __( 'Stacked', 'tribe' ), 'https://via.placeholder.com/100x60.png?text=Stacked' )
-					->add_option( self::LAYOUT_INLINE, __( 'Inline', 'tribe' ), 'https://via.placeholder.com/100x60.png?text=Inline' )
+					->add_option( __( 'Stacked', 'tribe' ), self::LAYOUT_STACKED, 'https://via.placeholder.com/100x60.png?text=Stacked' )
+					->add_option( __( 'Inline', 'tribe' ), self::LAYOUT_INLINE, 'https://via.placeholder.com/100x60.png?text=Inline' )
```
jbrinley, line 41: Thanks for catching this, @nealfennimore. We have some inconsistencies in Gutenblarghels that I'll clean up to help avoid this.

PR #447, opened 2020-04-15
Asset enqueues

jbrinley:
Work with the new `asset.php` build files for theme, admin, login, legacy, and editor assets.


wp-content/plugins/core/src/Assets/Admin/Scripts.php
```diff
@@ -0,0 +1,53 @@
...
+		foreach ( $this->build_parser->get_scripts() as $handle => $asset ) {
+			wp_register_script( $handle, $asset['uri'], $asset['dependencies'], $asset['version'], true );
+		}
+	}
+
+	/**
+	 * @action admin_enqueue_scripts
+	 */
+	public function enqueue_scripts(): void {
+
```
faction23, line 40: Very elegant ❤️

PR #455, opened 2020-04-22
Feeture/sqone 365/kitchen sink

dpellenwood:
Handles much of the default kitchen sink updates. 

* Updates sink styles based on new designs
* Separates t-content bits into separate t-sink & s-sink classes and partials for typography versus spacing styles.
* Updates a number of css classes used throughout to better align with naming structure.


wp-content/plugins/core/src/Panels/Types/Hero.php
```diff
@@ -17,8 +17,8 @@ class Hero extends Panel_Type_Config {
 	const FIELD_IMAGE                        = 'image';
 	const FIELD_CTA                          = 'cta';
 	const FIELD_TEXT_COLOR                   = 'text_color';
-	const FIELD_TEXT_LIGHT                   = 't-content--light';
-	const FIELD_TEXT_DARK                    = 't-content--dark';
+	const FIELD_TEXT_LIGHT                   = 't-theme--light';
+	const FIELD_TEXT_DARK                    = 't-theme--dark';
```
ryanurban, line 21: Nothing to update right now, just noting I'd like to probably migrate skins/themes to have their own prefixed namespace once we land on final approach.

wp-content/plugins/core/src/Templates/Controllers/Panel/Panel.php
```diff
@@ -90,7 +90,7 @@ protected function get_description( array $panel_vars ): string {
 	}
 
 	protected function get_description_classes( array $panel_vars ): array {
-		return [ 's-desc', 't-content' ];
+		return [ 's-desc', 's-sink t-sink' ];
```
ryanurban, line 93: Break these into single values?

wp-content/themes/core/assets/css/src/theme/color/_variables.pcss
```diff
@@ -6,48 +6,75 @@
...
 
 	/* -----------------------------------------------------------------------------
 	 * Backgrounds
 	 * ----------------------------------------------------------------------------- */
 
 	--color-bgd: var(--color-white);
 	--color-bgd-content: var(--color-white);
-	--color-bgd-login: #f1f1f1;
+	--color-bgd-login: var(--color-grey-light);
 	--color-bgd-selection: var(--color-primary);
```
ryanurban, line 69: Wondering if we need these here or at all. Ideally they should get ported up to a more appropriate vars partial?

wp-content/themes/core/assets/css/src/theme/color/_variables.pcss
```diff
@@ -6,48 +6,75 @@
...
 
 	/* -----------------------------------------------------------------------------
 	 * Links
 	 * ----------------------------------------------------------------------------- */
 
-	--color-link: var(--color-primary);
-	--color-link-hover: var(--color-secondary);
+	--color-link: var(--color-text);
+	--color-link-hover: var(--color-primary);
+	--color-link-focus: var(--color-primary-70);
```
ryanurban, line 77: Same thought, these should get hoisted up to sink vars

wp-content/themes/core/assets/css/src/theme/color/_variables.pcss
```diff
@@ -60,14 +87,22 @@
...
 
-	--color-border: var(--color-grey);
+	--color-border: var(--color-grey-light);
 
 	/* -----------------------------------------------------------------------------
-	 * Media: WP Oembed
+	 * Media
 	 * ----------------------------------------------------------------------------- */
 
 	--wp-oembed-bgd: var(--color-grey);
```
ryanurban, line 96: Same thought

wp-content/themes/core/assets/css/src/theme/color/_variables.pcss
```diff
@@ -60,14 +87,22 @@
...
 
+	/* -----------------------------------------------------------------------------
+	 * Alerts
+	 * ----------------------------------------------------------------------------- */
+
+	--color-alert-positive: rgba(13, 219, 145, 1);
+	--color-alert-negative: rgba(255, 61, 91, 1);
+	--color-alert-neutral: rgba(3, 134, 255, 1);
```
ryanurban, line 104: Same thought, where are these used?

wp-content/themes/core/assets/css/src/theme/spacing/sink.pcss
```diff
@@ -0,0 +1,153 @@
...
+		margin: var(--spacer-60) auto;
+	}
+
+	/* -----------------------------------------------------------------------------
+	 * Overline
+	 *
+	 * Overrides default selector spacing.
+	 * ----------------------------------------------------------------------------- */
+
+	.t-overline {
```
ryanurban, line 146: What is this, different from .h1-h6?

wp-content/themes/core/assets/css/src/theme/typography/_mixins.pcss
```diff
@@ -5,32 +5,202 @@
...
+/* -------------------------------------------------------------------------
+ * Label
+ * ------------------------------------------------------------------------- */
+
+@define-mixin t-label {
+	font-family: var(--font-family-sans-serif);
+	font-size: 14px;
+	line-height: 1.4;
+	font-weight: 400;
+}
```
ryanurban, line 177: These probably belong in forms mixins?

wp-content/themes/core/assets/css/src/theme/typography/_mixins.pcss
```diff
@@ -5,32 +5,202 @@
...
+ * Text Button/Call To Action
+ * ------------------------------------------------------------------------- */
+
+@define-mixin t-text-button {
+	font-family: var(--font-family-sans-serif);
+	font-size: 18px;
+	line-height: 1.5556;
+	font-weight: 700;
+	letter-spacing: 0.1111px;
+}
```
ryanurban, line 155: Probably belong in buttons mixins?

wp-content/themes/core/assets/css/src/theme/typography/_utilities.pcss
```diff
@@ -1,66 +1,125 @@
...
+	@mixin t-input;
+}
+
+/* -------------------------------------------------------------------------
+ * Label
+ * ------------------------------------------------------------------------- */
 
+.t-label {
+	@mixin t-label;
 }
```
ryanurban, line 125: Same q as from mixins

wp-content/themes/core/assets/css/src/theme/typography/_utilities.pcss
```diff
@@ -1,66 +1,125 @@
...
- *     ...
- *   </div>
- * </div>
- *
- * Example:
- * <div class="t-content t-content--light">
- *   ...
- * </div>
- *
+ * Display X-Large
```
ryanurban, line 8: Nothing to change, noting for future to follow up on this, wondering if it's valuable or not to have all of these as helper classes that then output a bunch of CSS, especially if we are going to 99% of the time us our sink helpers as these won't get us much mileage without responsive handling and as of now we both agree we don't want to go down that wormhole.

wp-content/themes/core/assets/css/src/theme/typography/_variables.pcss
```diff
@@ -6,43 +6,25 @@
...
-	--line-height-s: 1.25;
-	--line-height-ss: 1.15;
-	--line-height-even: 1;
+	--font-size-body-large: 20px;
+	--font-size-body: 16px;
+	--font-size-body-small: 12px;
 
 	/* -----------------------------------------------------------------------------
 	 * Border Radius
+	 * TODO: Where do these belong?
```
ryanurban, line 27: Good question, have the same q for like border colors too

wp-content/themes/core/assets/css/src/theme/typography/all.pcss
```diff
@@ -15,3 +15,6 @@
 @import "p.pcss";
 @import "quotes.pcss";
 @import "tables.pcss";
+
+/* Themes */
+@import "themes.pcss";
```
ryanurban, line 20: Nothing to change, for the future, this should get rolled up into whatever theme/skin approach we land on

wp-content/themes/core/assets/css/src/theme/typography/lists.pcss
```diff
@@ -5,84 +5,108 @@
...
+li {}
 
 /* -------------------------------------------------------------------------
- * UL
+ * Unordered List
  * ------------------------------------------------------------------------- */
 
-.list-bullet,
-.t-content ul {
+.t-list--bullets,
```
ryanurban, line 30: have one extra dash, should be `.t-list-bullets`

wp-content/themes/core/components/document/document.twig
```diff
@@ -7,8 +7,7 @@
 
 		{{ do_action( 'tribe/body_opening_tag') }}
 
-		{# TODO: Move related a11y-skip-link css partial to this component #}
-		<a href="#a11y-skip-link-content" class="a11y-skip-link a11y-visual-hide">{{ __( 'Skip to main content' )|esc_html }}</a>
+		<a href="#a11y-skip-link-content" class="a11y-skip-link u-visual-hide">{{ __( 'Skip to main content' )|esc_html }}</a>
```
ryanurban, line 10: When we get to a11y, we'll want add this as the first child item in site header/nav wrapper. Solves a "all things must be wrapped in a role based thing" issue.

wp-content/themes/core/components/tabs/Tabs.php
```diff
@@ -88,7 +88,7 @@ class Tabs extends Context {
 			self::DEFAULT => [],
 		],
 		self::TAB_CONTENT_CLASSES       => [
-			self::DEFAULT       => [ 'c-tab__content', 't-content' ],
+			self::DEFAULT       => [ 'c-tab__content', 's-sink t-sink' ],
```
ryanurban, line 91: break into separate values?

wp-content/themes/core/components/video/css/video.pcss
```diff
@@ -4,11 +4,15 @@
  *
  * ----------------------------------------------------------------------------- */
 
+:root {
+	--color-video-trigger-overlay: linear-gradient(360deg, #222 0%, rgba(34, 34, 34, 0.1) 54.52%, rgba(34, 34, 34, 0) 100%);
+}
```
ryanurban, line 9: assuming this remains to be done and will be handled as part of your work today, but we should have a variables partial for this component : )

wp-content/plugins/core/src/Panels/Types/Hero.php
dpellenwood: :+1: Agreed.

wp-content/plugins/core/src/Templates/Controllers/Panel/Panel.php
dpellenwood: Good catch. Fixing...

wp-content/themes/core/assets/css/src/theme/color/_variables.pcss
dpellenwood: These make sense to move to their respective partial, yes. Will do.

wp-content/themes/core/assets/css/src/theme/typography/lists.pcss
dpellenwood: I see `bullets` as a modifier for the `list` block, hence the double dash. List could have other marker styles besides bullets, couldn't they?

wp-content/themes/core/components/document/document.twig
dpellenwood: Good thought. I'll add a TODO there.

wp-content/themes/core/components/video/css/video.pcss
dpellenwood: Yeppers. :100:

wp-content/themes/core/assets/css/src/theme/typography/lists.pcss
ryanurban: Hmm, true, I suppose these would be flavors of lists, which if that's the case, maybe it's weird that we use modifiers when they wouldn't necessarily build on one another and are instead straight up versions of a list (`.t-list-bullets` or `.t-list-styled`), it's not like we have a base `.t-list`, though perhaps we should?

wp-content/themes/core/assets/css/src/theme/color/_variables.pcss
dpellenwood: These are unused at this time. They are part of design's new sink. I just added them here until we get to a place/component where they will be used. (Forms are an example, I think.)

wp-content/themes/core/assets/css/src/theme/spacing/sink.pcss
dpellenwood: It's a style that can be applied to any element, such as a `p` that appears before a heading: http://p.tri.be/Zqe47q

wp-content/themes/core/assets/css/src/theme/typography/_mixins.pcss
dpellenwood: That's one way to think about it. However, design uses these typographic styles throughout the designs (such as the stylized list numbering) which means that don't just apply to forms.

wp-content/themes/core/assets/css/src/theme/typography/_mixins.pcss
dpellenwood: I think it makes more sense to keep all the typography mixins with the sink typography. As again, these may be used for other "components" than just buttons.

wp-content/themes/core/assets/css/src/theme/typography/_utilities.pcss
dpellenwood: Hmm... design tends to use these "base" typographic styles many time throughout the designs where there is no responsive treatment. I've found this system super helpful on the last few projects. Happy to reevaluate as we proceed, but for now, I'd like to keep them available.

wp-content/themes/core/assets/css/src/theme/typography/_variables.pcss
dpellenwood: Yup. Thinking that this will maybe end up landing in a  theme/decorations directory eventually. I see some box-shadows stuff in the designs as well now.

wp-content/themes/core/assets/css/src/theme/typography/all.pcss
dpellenwood: :+1:

wp-content/themes/core/assets/css/src/theme/color/_variables.pcss
ryanurban: 👍

wp-content/themes/core/assets/css/src/theme/typography/_utilities.pcss
ryanurban: 👍

wp-content/themes/core/assets/css/src/theme/typography/_mixins.pcss
ryanurban: Hmm, don't love the contextual naming of these typographic things (which I see is coming from figma) as it implies...context. I'll stick a note in this to follow up with design on for later, but I see what you're saying and based on how it's setup, agreed.

wp-content/themes/core/assets/css/src/theme/typography/_mixins.pcss
ryanurban: Sticking a pin in this for later, carry on, part of the contextual naming note.

wp-content/themes/core/assets/css/src/theme/spacing/sink.pcss
ryanurban: Got it, does that then also mean that all the other context named typo helper classes should also get added to here?

dpellenwood: They are already available via this sink spacing file and the typography sink styles (as is this typographic style). This code is just to handle an edge case/exception for spacing via the sibling selector.

wp-content/themes/core/assets/css/src/theme/typography/lists.pcss
dpellenwood: Yeah, I wonder about that, too. I'll go ahead and add a `t-list` selector to the shared styles block at the top of the file for consistancy.

wp-content/themes/core/assets/css/src/theme/spacing/sink.pcss
ryanurban: Talking about all of these `.t-display-x-large`, similar to `.t-overline`, it doesn't look like we are applying any vertical spacing to them by default when wrapped in our vert spacing helper. Confirming that is intentional and gtg?

ryanurban: Perhaps we could key in on some global defaults with attribute selectors `[class*="t-display"]` and do by groupings

dpellenwood: Confirmed. That is intentional and g2g. From the designs, the overline style is unique in that it can be applied to other elements to produce a different spacing flow than the other blocks/tags. I don't think we want to apply default sink spacing to all the typographic helpers out of the box. That would land us back in the over-use of `t-content` scenario where _all_ the typograhic styles have automatic spacing which needs to be overridden.

PR #457, opened 2020-04-23
Feeture/sqone 173/kitchen sink media

dpellenwood:
Handles base grid & kitchen sink images.  Ticketing Audio & Video spearately b/c they will require some planning and fixes for oEmbeds.

Also disables the default block theme styles from rendering on the FE.


wp-content/themes/core/assets/css/src/theme/grid/_variables.pcss
```diff
@@ -9,20 +9,49 @@
...
+	--grid-margin-small: 24px;
+	--grid-gutter-small: 24px;
+	--grid-margin-small-negative: calc(var(--grid-margin-small) * -1);
 	--grid-gutter-small-negative: calc(var(--grid-gutter-small) * -1);
 	--grid-gutter-small-half: calc(var(--grid-gutter-small) / 2);
 	--grid-gutter-small-half-negative: calc(var(--grid-gutter-small-half) * -1);
 
-	/* Widths */
-	--grid-width: calc(var(--grid-width-default) + var(--grid-gutter) * 2);
+	/* Grid Column Max Widths */
```
ryanurban, line 34: I can't see ever using these, we don't actually do a traditional grid system? 😀

dpellenwood: Actually, I use them all the time, and have used them here. When design requests a specific width for, say an aligned image in the grid: http://p.tri.be/jQFTHd, this allows you to set the max-width to 5 columns (consistently) rather than having to recall what the pixel value for 5 columns is.

ryanurban: Hah, well alrighty then. If that's the case, we should also work it into the grid as if it's only in these vars I'm sure very few people will use them.

ryanurban: Can stick a pin in that for grid/layout overhaul time

wp-content/plugins/core/src/Assets/Theme/Styles.php
```diff
@@ -37,4 +37,13 @@ public function enqueue_styles(): void {
...
 		}
 	}
+
+	/**
+	 * Dequeue WP Core's default block styles from the public FE.
+	 *
+	 * @action wp_enqueue_scripts
+	 */
+	public function dequeue_block_styles(): void {
+		wp_dequeue_style( 'wp-block-library' );
```
jbrinley, line 47: I need to look at this a little more. My understanding was that core block styles were an opt-in feature. I'll get back to you this afternoon.

jbrinley: I was mistaken. There are two separate stylesheets: `wp-block-library` and `wp-block-library-theme`. Only the latter is controlled by the `add_theme_support( 'wp-block-styles' )` directive.

With that said, are you sure you really want to dequeue this, @dpellenwood? What's in there is pretty theme agnostic, and I suspect we'll have more effort replacing it than we would overriding the pieces we don't want.

dpellenwood: Thanks for checking, Jonathan. Yes, I'm sure. It requires additional code to undo some of the layout & spacing values applied in that file to work with our css/design system. I know we'll end up porting over some of the styles; but, much of it we already support in our system in a different way.

wp-content/plugins/core/src/Theme/Theme_Subscriber.php
```diff
@@ -29,7 +29,7 @@ private function media(): void {
 		$this->image_wrap();
 		$this->image_links();
 		$this->disable_responsive_images();
-		$this->oembed();
+		//$this->oembed();
```
dpellenwood, line 32: @jbrinley , just want to point out that I have disabled our custom oembed filter for the time being so default embed can work.

wp-content/themes/core/assets/css/src/theme/media/wp-audio-video.pcss
```diff
@@ -18,3 +22,53 @@
 		background-image: none;
 	}
 }
+
+/* -----------------------------------------------------------------------------
```
dpellenwood, line 26: @ryanurban just pointing out these styles to support WordPress block editor oEmbeds while we suss out our approach for custom stuff.

PR #458, opened 2020-04-27
Block overrides

jbrinley:
Based on the requirements from https://docs.google.com/document/d/1VzRL6VrLadKsHEGogo8wQCwxH0rFrY5HD_XpLukQqRY/edit#heading=h.me9uedb1p9r

* Disables the core "Buttons" block
* Adds our custom "Button" block
* Sets the available font sizes to a single option, "Lead-In", with the identifier "t-body-large"
* Adds the block style "Overline" to core paragraph and header blocks


wp-content/plugins/core/src/Blocks/Types/Button.php
```diff
@@ -0,0 +1,45 @@
...
+		return $this->factory->content()->section()
+			->add_field(
+				$this->factory->content()->field()->link( self::LINK )
+					->set_label( __( 'Link', 'tribe' ) )
+					->build()
+			)
+			->add_field(
+				$this->factory->content()->field()->text( self::ARIA )
+					->set_label( __( 'Accessibility Label', 'tribe' ) )
+					->set_placeholder( __( 'enter a label for screen readers', 'tribe' ) )
```
dpellenwood, line 39: Super minor, but wondering if we should make this a little more plain-language? I.e - `Provide an optional label for screen readers.`

wp-content/plugins/core/src/Blocks/Types/Button.php
```diff
@@ -0,0 +1,45 @@
...
+
+	public const NAME = 'tribe/button';
+
+	public const LINK = 'link';
+	public const ARIA = 'aria_label';
+
+	public function build(): Block_Type_Interface {
+		return $this->factory->block( self::NAME )
+			->set_label( __( 'Button', 'tribe' ) )
+			->set_dashicon( 'align-center' )
```
dpellenwood, line 20: Will we need a custom icon for our button block?

wp-content/plugins/core/src/Templates/Controllers/Block/Button.php
```diff
@@ -0,0 +1,32 @@
...
+
+class Button extends Block_Controller {
+
+	public function render( string $path = '' ): string {
+		$link = $this->get_link( $this->attributes );
+
+		return $this->factory->get( Link::class, [
+			Link::URL        => $link['url'],
+			Link::CONTENT    => $link['text'],
+			Link::TARGET     => $link['target'],
```
dpellenwood, line 17: Only because I'm ignorant, does link target get populated?

wp-content/plugins/core/src/Blocks/Types/Button.php
jbrinley: Yes, we will need a custom icon. I'm waiting on Sam to set up a framework for that.

wp-content/plugins/core/src/Templates/Controllers/Block/Button.php
jbrinley: There's a UI ticket to add that.

PR #459, opened 2020-04-27
Feeture/sqone 133/load webfonts via css

dpellenwood:
Updates Sq1 to load web fonts via CSS rather than JS.

[SQONE-133]

[SQONE-133]: https://moderntribe.atlassian.net/browse/SQONE-133

### DIFF
--- a/wp-content/plugins/core/src/Theme/Config/Web_Fonts.php
+++ b/wp-content/plugins/core/src/Theme/Config/Web_Fonts.php
@@ -100,7 +100,7 @@ class Web_Fonts {
 	}
 
 	/**
-	 * Returns a fully-qualified Typekit font kit URL from the Project ID value.
+	 * Returns a fully-qualified Google Fonts URL from an array of font keys.
 	 *
 	 * @return string
 	 */
--- a/wp-content/plugins/core/src/Theme/Config/Web_Fonts.php
+++ b/wp-content/plugins/core/src/Theme/Config/Web_Fonts.php
@@ -12,24 +12,12 @@ class Web_Fonts {
 	private const GOOGLE_API       = 'https://fonts.googleapis.com';
 
 	/**
-	 * @var string
-	 */
-	private $plugin_file;
-
-	/**
 	 * @var array
 	 */
 	private $fonts;
 
-	/**
-	 * @var array
-	 */
-	private $font_urls = [];
-
-	public function __construct( string $plugin_file, array $fonts = [] ) {
-		$this->plugin_file = $plugin_file;
+	public function __construct( array $fonts = [] ) {
 		$this->fonts       = $fonts;
-		$this->set_font_urls();
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


wp-content/plugins/core/src/Theme/Config/Web_Fonts.php
```diff
@@ -20,115 +21,94 @@ class Web_Fonts {
...
+	private function get_typekit_url(): string {
+		if ( empty( $this->fonts[ self::PROVIDER_TYPEKIT ] ) ) {
+			return '';
+		}
+
+		return sprintf( '%s%s.css', trailingslashit( self::TYPEKIT_API ), $this->fonts[ self::PROVIDER_TYPEKIT ] );
+	}
+
+	/**
+	 * Returns a fully-qualified Typekit font kit URL from the Project ID value.
```
ryanurban, line 103: Misspelled Google 😂

wp-content/plugins/core/src/Theme/Theme_Definer.php
```diff
@@ -106,25 +106,39 @@ public function define(): array {
...
+			 * @var string CSS file URL.
+			 *
+			 * A complete URL to a css file containing custom @font-face definitions.
+			 * Useful for other 3rd-party web font providers such as fonts.com.
+			 *
+			 * Example: `'https://fast.fonts.net/cssapi/e6dc9b99-64fe-4292-ad98-6974f93cd2a2.css'`
+			 *
 			 */
-			self::CONFIG_CUSTOM_FONTS => [],
+			self::CONFIG_CUSTOM_FONTS => '',
```
jbrinley, line 135: Is it possible that we might have multiple sources for custom fonts? I.e., should this be an array of URLs, or just a single string?

wp-content/plugins/core/src/Theme/Config/Web_Fonts.php
```diff
@@ -5,130 +5,101 @@
...
+	 * Visual Editor Fonts
 	 *
-	 * @action wp_head
-	 * @action login_head
+	 * @action after_setup_theme
 	 */
-	public function load_fonts() {
+	public function add_visual_editor_fonts(): void {
+		foreach( $this->get_font_urls() as $url ) {
+			add_editor_style( $url );
```
jbrinley, line 53: This is only going to load styles for the MCE editor, not the block editor. Do we need these fonts in the block editor, too?

wp-content/plugins/core/src/Theme/Theme_Definer.php
dpellenwood: Good question, I wasn't planning on multiple; but after discussion w/ FE, we decided it's not a bad idea to make it support multiple. This has been updated.

wp-content/plugins/core/src/Theme/Config/Web_Fonts.php
dpellenwood: Good catch. I though the add_editor_style was working for both. It was not. I've updated to accommodate.

PR #461, opened 2020-04-27
Feeture/SQONE 176/forms

jamiepaul:
JIRA: https://moderntribe.atlassian.net/browse/SQONE-176

**WORK IN PROGRESS**

### DIFF
--- a/wp-content/themes/core/assets/css/src/theme/forms/_mixins.pcss
+++ b/wp-content/themes/core/assets/css/src/theme/forms/_mixins.pcss
@@ -76,8 +76,8 @@
 @define-mixin form-control {
 	transition: var(--transition);
 	padding: var(--form-control-padding);
-	border: var(--form-control-border-size) solid var(--form-control-border-color);
-	border-radius: var(--form-control-border-radius);
+	border: var(--form-border-size) solid var(--form-border-color);
+	border-radius: var(--form-border-radius);
 	background-color: var(--form-control-bgd-color);
 	display: block;
 	width: 100%;
@@ -104,7 +104,7 @@
 	&:active {
 
 		&:after {
-			background-color: var(--form-control-border-color-focus);
+			background-color: var(--form-border-color-focus);
 		}
 	}
 
@@ -113,20 +113,20 @@
 
 		display: block;
 		transition: var(--transition);
-		content: '\e900';
+		content: var(--icon-chevron-down);
 		text-align: center;
 		height: calc(var(--form-control-height) - 2px);
 		width: var(--form-control-select-ui-width);
 		line-height: var(--form-control-height);
-		background-color: var(--form-control-border-color);
+		background-color: var(--form-border-color);
 		color: var(--form-control-select-ui-color);
 		position: absolute;
 		font-size: 25px;
 		right: 1px;
 		top: 1px;
 		pointer-events: none;
-		border-top-right-radius: var(--form-control-border-radius);
-		border-bottom-right-radius: var(--form-control-border-radius);
+		border-top-right-radius: var(--form-border-radius);
+		border-bottom-right-radius: var(--form-border-radius);
 	}
 
 	select {
@@ -230,7 +230,7 @@
 	label {
 
 		&:before {
-			border-radius: var(--form-control-border-radius);
+			border-radius: var(--form-border-radius);
 		}
 
 		&:after {
@@ -281,19 +281,19 @@
 	position: relative;
 
 	input {
-		padding-right: var(--form-control-icon-padding-right);
+		padding-right: var(--form-control-height);
 	}
 
 	&:before {
 		position: absolute;
 		right: var(--form-control-icon-position-right);
 		top: var(--form-control-icon-position-top);
-		color: var(--form-control-icon-color);
+		color: lightgrey;
 		height: 100%;
-		width: var(--form-control-icon-width);
-		line-height: var(--form-control-icon-line-height);
+		width: var(--form-control-height);
+		line-height: var(--form-control-height);
 		text-align: center;
 		z-index: 1;
-		font-size: var(--form-control-icon-font-size);
+		font-size: 20px;
 	}
 }
--- a/wp-content/themes/core/assets/css/src/theme/forms/_mixins.pcss
+++ b/wp-content/themes/core/assets/css/src/theme/forms/_mixins.pcss
@@ -17,13 +17,29 @@
 
 /* -----------------------------------------------------------------------------
  *
+ * Form Label
+ *
+ * ----------------------------------------------------------------------------- */
+
+@define-mixin form-label {
+	display: block;
+	margin-bottom: var(--form-margin-small);
+	padding: 0;
+	color: var(--form-color);
+	letter-spacing: normal;
+	text-transform: none;
+	font: normal 600 var(--form-font-size-secondary) / 1.4 var(--form-font-family);
+}
+
+/* -----------------------------------------------------------------------------
+ *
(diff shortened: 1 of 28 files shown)


wp-content/themes/core/assets/css/src/theme/forms/_mixins.pcss
```diff
@@ -11,8 +11,8 @@
  * ----------------------------------------------------------------------------- */
 
 @define-mixin form-control-group-radio-checkbox {
-	padding-top: var(--form-control-group-radio-checkbox-padding-top);
-	padding-bottom: var(--form-control-group-radio-checkbox-padding-bottom);
+	padding-top: 8px;
```
dpellenwood, line 14: We have spacing vars available (at least for 8px, `--spacer-10`). I think it might be worth it to use that?

wp-content/themes/core/assets/css/src/theme/forms/_mixins.pcss
```diff
@@ -22,12 +22,12 @@
...
-	margin-top: var(--form-control-helper-margin-top);
-	padding: var(--form-control-helper-padding);
-	color: var(--form-control-helper-color);
+	margin-top: 8px;
+	padding: 0 8px;
+	color: var(--form-color-secondary);
 	letter-spacing: var(--form-control-helper-letter-spacing);
 	text-transform: var(--form-control-helper-text-transform);
-	font: var(--form-control-helper-font-style) var(--form-control-helper-font-weight) var(--form-control-helper-font-size) / var(--form-control-helper-font-line-height) var(--form-control-helper-font-family);
+	font: normal var(--form-control-helper-font-weight) var(--form-control-helper-font-size) / var(--form-control-helper-font-line-height) var(--form-control-helper-font-family);
```
dpellenwood, line 30: The shorthand font declaration here doesn't always work x-browser. This is due to the keyword `inherit` occasionally being used for line-height, which is [invalid per the spec](https://stackoverflow.com/a/5518592) for shorthand properties, like `font`. The resolution I have generally used on projects it to just break up the shorthand `font` declaration into separate properties for each.  This is true throughout the forms sink (and SquareOne in general).

wp-content/themes/core/assets/css/src/theme/forms/_mixins.pcss
```diff
@@ -111,20 +111,20 @@
 
 		display: block;
 		transition: var(--transition);
-		content: var(--form-control-select-ui-icon);
+		content: '\e900';
```
dpellenwood, line 114: I think we should still use an icon var value here (`--icon-chevron-down`), since the actual character code may change depending on the structure of the icon font.

wp-content/themes/core/assets/css/src/theme/forms/_mixins.pcss
```diff
@@ -156,31 +156,31 @@
 	}
 
 	label {
-		color: var(--form-control-radio-checkbox-label-color);
```
dpellenwood, line 159: @ryanurban , I wonder if we should leverage the typography mixin for labels here?

ryanurban: Ideally yes, for sure, for the actual typographic things

wp-content/themes/core/assets/css/src/theme/forms/_variables.pcss
```diff
@@ -76,21 +53,22 @@
...
-	--form-control-border-color: var(--form-border-color);
-	--form-control-border-color-focus: var(--form-border-color-focus);
-	--form-control-border-radius: var(--form-border-radius);
-	--form-control-border-size: var(--form-border-size);
-	--form-control-border-size-focus: var(--form-border-size-focus);
+	--form-control-border-color: var(--color-neutral-20);
+	--form-control-border-color-focus: var(--color-neutral-90);
+	--form-control-border-radius: 0;
+	--form-control-border-size: 1px;
+	--form-control-border-size-focus: var(--form-control-border-size);
```
ryanurban, line 64: I think my preference would be to keep the global set of `--form-border-`vars up top and nix these control & field vars in favor of using those. Thoughts? It's nice to have sets of vars segmented into their groupings for both organization but clarity too.

wp-content/themes/core/assets/css/src/theme/forms/_variables.pcss
```diff
@@ -5,68 +5,45 @@
...
-	 * UI
-	 * ----------------------------------------------------------------------------- */
-
-	--form-ui-icon-color: lightgrey; /* Control icons */
-	--form-ui-icon-font-size: 20px; /* Control icons */
-	--form-ui-color: var(--color-grey-dark); /* Select icon */
-	--form-ui-bgd-color: var(--form-border-color);  /* Select, Checkbox, & Radio */
-	--form-ui-bgd-color-focus: var(--form-border-color-focus);  /* Select */
-	--form-ui-icon-select: '\e900'; /* Select */
-	--form-ui-icon-select-font-size: 25px; /* Select */
```
ryanurban, line 39: In going through and scoping out the clean up, as well as some of the lingering issues from my first pass, I'm noticing that there is a global set of UI based vars that would be nice to have/use over the more granular vars, a great example is sticking with `--form-ui-bgd-color` vs. `--form-control-bgd-color` and cleaning up the latter. I would imagine we'll have a whole global set of stuff like this that would be lovely to standardize into a set of UI based vars.

wp-content/themes/core/assets/css/src/theme/forms/_mixins.pcss
jamiepaul: absolutely, hadn't made a pass looking for where we can use vars yet but that's on my list!

wp-content/themes/core/assets/css/src/theme/forms/_mixins.pcss
jamiepaul: good insight, I can break those up across the board for forms

wp-content/themes/core/assets/css/src/theme/forms/_mixins.pcss
jamiepaul: 👍 have not done a vars pass yet but will add the appropriate chevron var there. For some reason, we had that as the value for `--form-control-select-ui-icon` before I removed it.

wp-content/themes/core/assets/css/src/theme/forms/_mixins.pcss
jamiepaul: will update 👍

wp-content/themes/core/assets/css/src/theme/forms/_variables.pcss
jamiepaul: I was 50/50 on keeping them just as `form-border--` up top, or adding them with the controls. Since they were used for controls, I figured it was logical to add them w/ controls BUT it's less to type if we just keep them as `form-border--*` up top, I'm in favor of that

wp-content/themes/core/assets/css/src/theme/forms/_variables.pcss
jamiepaul: @ryanurban @dpellenwood Let's chat through this. I feel that the global UI vars are too general and it's hard to recall where each is used. For me, this is an area that contributes to the "nesting" of forms vars that makes it tedious to update on project setup. Many in this set were only used in 1-2 spots.

ryanurban:
@jamiepaul heads up that you should wrap up the forms work in `feeture/forms` 🙃

PR #468, opened 2020-05-08
Feeture/sqone 182/editor formats

dpellenwood:
Handles a number of block editor styles, last round of kitchen sink work, text formats, and a bunch of other sink shenanigans. 

[SQONE-182]

[SQONE-182]: https://moderntribe.atlassian.net/browse/SQONE-182


gulp_tasks/postcss.js
```diff
@@ -5,10 +5,31 @@ const rename = require( 'gulp-rename' );
 const gulpif = require( 'gulp-if' );
 const concat = require( 'gulp-concat' );
 const browserSync = require( 'browser-sync' );
-const PrefixWrap = require( 'postcss-prefixwrap' );
 const postcssFunctions = require( '../dev_components/theme/pcss/functions' );
 const pkg = require( '../package.json' );
 
+const selectorsBefore = [
```
faction23, line 11: You foresee these being pretty stable down the road right, like not something we need to manage as we add more custom blocks?

dpellenwood: Yes. We won't have to target anything custom. These are just to transform our kitchen sink selectors to match the WP Core blocks. In fact, I'll remove the alignment ones. The markup between the editor and FE doesn't line up for these.

faction23: sweet

PR #470, opened 2020-05-08
Require `so` global docker

jbrinley:
The `so` global docker environment should become our canonical tool for interacting with Docker and the project containers.

* Removes the global docker directory from SquareOne
* Points all npm scripts to `so` instead of bash scripts
* Removes all the bash scripts that can be accomplished with `so`
* Adds a `squareone.yml` config file


squareone.yml
```diff
@@ -0,0 +1,6 @@
+# SquareOne Project Docker Config
```
defunctl, line 1: Let's add a bit more clarity here about how this can be customized and why.

e.g. something like: run `so config:copy` and move the `/home/$USER/.config/squareone/squareone.yml` to override this file to fully customize docker for your project.

dev/docker/local_domain_migration.sh
```diff
@@ -29,5 +29,4 @@ ${D_COMMAND} exec tribe-mysql \
  mysql -uroot -ppassword $DB_NAME \
  -Ne "UPDATE ${DB_PREFIX}options SET option_value=REPLACE( 'option_value', '$SOURCE_DOMAIN', '$TARGET_DOMAIN' ) WHERE option_name='siteurl'"
 
-/bin/bash $SCRIPTDIR/wp.sh \
- search-replace "$SOURCE_DOMAIN" "$TARGET_DOMAIN" --all-tables-with-prefix --verbose
+so wp -- search-replace "$SOURCE_DOMAIN" "$TARGET_DOMAIN" --all-tables-with-prefix --verbose
```
defunctl, line 32: We can remove this file if the `so` global implementation is working for you.

PR #473, opened 2020-05-13
Feeture/build to theme

faction23:
Like it says in the title ;) I'll be updating docs later this week.


jbrinley:
@marcossantiago: how does the `/script` directory relate to the `/dev/deploy` directory? Does it replace it? Can we combine them? Can these scripts live in the `/dev` directory instead of the project root?

marcossantiago:
@jbrinley Nothing is set on stone, we can setup in either way but the main idea is to use these set scripts for every project like square-one, devops, infra, non-wordpress web projects and since many projects do not have the `dev` folder, it would make more sense to stay on the root folder. However, I have no problems in putting them into `/dev/`. 

The `./dev/deploy` would be used to host the configuration files related to the servers. For example, WPEngine configuration repositories.

jbrinley:
Are these scripts intended to vary by project, or would all projects using them need to stay in sync?

marcossantiago:
Yes, they are intended to vary by project. The idea is to have a normalized bootstrap process that it's language agnostic, so each project can use have their own commands, scripts, processes but still being consistent through all projects. I forgot to add a `test` script.

nickpelton:
Hey Marcos, deploy stuff looks good. Any reason we can't put all these files into the /dev/deploy folder? Even the Jenkins file can be put in there to keep things tidy.

PR #476, opened 2020-05-14
Simplify and align Content Block component with intended usages.

dpellenwood:
Cleans up & standardizes the content block component a bit.


wp-content/themes/core/components/content-block/content-block.twig
```diff
@@ -0,0 +1,23 @@
+{# Component: Content Block #}
+
+<{{ tag }} {{ classes }} {{ attrs }}>
+	<div {{ container_classes }}>
+
+		{% if leadin %}
+			{{ leadin }}
+		{% endif %}
```
jbrinley, line 8: Do we need all these `if`s?

dpellenwood: Nope, we do not. Removed.

PR #482, opened 2020-05-19
Remove panel builder

jbrinley:
We're not planning to support panel builder with the new version of Square One. This PR removes the plugin and all of the integration code (registration of default panels, template controllers, etc.).


jbrinley:
@dpellenwood, I've left in place the components, because it looks like we're using the same components for blocks. Should some/all of them be removed?

dpellenwood:
Woot! Thanks for cleaning this up, Jonathan. Yes, we are "converting" the existing panel templates into block usage. I've [created a ticket](https://moderntribe.atlassian.net/browse/SQONE-413) to make a pass at the remaining legacy panels code in the theme.

PR #488, opened 2020-06-03
Feeture/sqone 224/panel interstitial

dpellenwood:
FF: Builds out the Interstitial Panel


dpellenwood:
@faction23, the Interstitial & Hero panels are effectively the same markup & base styles. The Hero just get sa couple extra fields (lead-in & description), and has different vertical spacing. Therefore, I'd like to try and make the templates, controllers, & CSS as similar as possible.

faction23:
> @faction23, the Interstitial & Hero panels are effectively the same markup & base styles. The Hero just get sa couple extra fields (lead-in & description), and has different vertical spacing. Therefore, I'd like to try and make the templates, controllers, & CSS as similar as possible.

Noted!

PR #492, opened 2020-06-07
Feeture/squareone docker image

jbrinley:
* Replaces the included PHP Dockerfile with a pull from the *new* [Docker Hub-hosted PHP image](https://hub.docker.com/r/moderntribe/squareone-php).
* Removes services from `docker-compose.yml` that are not needed on typical projects.
* Adds documentation for maintaining/modifying `docker-compose.yml`.
* Adds service recipes for adding atypical services to `docker-compose.yml` for a project.


jbrinley:
Request for comment. I would like to host our base PHP docker image(s) on Docker Hub. https://hub.docker.com/repository/docker/moderntribe/squareone-php

Why?
* We base our image on phpdocker.io, which does not tag versions. If you and I have different versions of the base image cached in our local docker environments, we will get different results when we build our local containers. This inconsistency has given us a number of difficult to diagnose issues over the years.
* Hosting the image on Docker Hub speeds up the initial boot of a dev environment, both locally and in automated tests running on GH actions.
* We have trouble getting our dev team to properly tag and update project-specific images. This sometimes leads to projects trying to build conflicting containers. By separating the base container, the override process becomes a more distinct step that encourages good practice.

Concerns:
* "We sometimes need to override the PHP container for projects" - Yes we do, and we can still do that. I've documented how in this PR.
* "I don't know how to update these images" - It's documented in this PR.
* Anything else to be concerned about?

jbrinley:
Thanks for the thoughts, @lucatume

> 1. https://www.fromlatest.io/#/ -- I've used this to validate my Dockerfiles

Thanks. I've run our new Dockerfile through that, and it's not reporting anything unexpected (I'll note that it has trouble reading commands that span multiple lines).

> 2. I do not see, in the `docker-compose.yml` file, any reference to `USER`.

I don't have a good understanding of what's necessary to make this work on Linux. I'll leave that to @defunctl to comment on. My presumption has been that it is something reasonably handled with a local `docker-compose.override.yml`, but I could easily be mistaken.

> 3. I have a question about the use of the stack in CI: does that require booting the global stack too? 

We do spin up a subset of the global containers for tests (see https://github.com/moderntribe/square-one/blob/epic/fidgety-feet/dev/docker/docker-compose.gh.yml). There's a probably a way to optimize that, but it works reasonably well as-is.

defunctl:
I think this is a fantastic idea. 

Perhaps some additional documentation around something that should be submitted to the core image @ https://github.com/moderntribe/squareone-docker-images in the event something should go out to everyone like if a core extension was missed.

Regarding the permissions issue, thanks for posting that @Luc45, I'll definitely investigate, SquareOne Global Docker could certainly set the proper user/group environment variable ID's.

defunctl:
Not to derail this conversation too much, but what's happening in Tric is only working because it's based off a `php` image which takes these variables into account in their Dockerfile specific to Apache: https://github.com/docker-library/php/blob/master/apache-Dockerfile-block-1#L14 and https://github.com/docker-library/php/blob/master/7.2/buster/apache/Dockerfile#L65

This doesn't work with phpdocker.io because it can't start php-fpm as non root user and can't write to `/run` etc...

defunctl:
I've got a couple of ideas for the permissions issues, I'll submit some PR's and continue the discussion there.

lucatume:
@defunctl to handle the file permission issues, in `tric` and other custom-built stacks, I've used the specific image-provided support (as is the case for the Apache version of the WordPress one) or a user redirection using [boatbox/fixuid(https://github.com/boxboat/fixuid).

There are other, similar, solutions, but that got me far enough for my needs.

PR #493, opened 2020-06-10
Add SQ1_UID docker argument and environment variable

defunctl:
## What does this do/fix?

This should hopefully allow SquareOne Global Docker to set the host's ID as the `SQ1_UID` environment variable and pass it to docker-compose.yml to pass to the Dockerfile as an argument (untested).

## QA

https://github.com/moderntribe/squareone-docker-images/pull/1#discussion_r437787226

## Tests

Does this have tests?

- [ ] Yes
- [X] No, this doesn't need tests because...it's a config file :)


defunctl:
This is no longer needed, will address this in square one global docker.

PR #501, opened 2020-07-01
A move to MVC/MCC

speerface:
## What does this do/fix?

A rather large roadblock we've encountered when handing off projects to clients is the fairly complex templating system we use. Our mix of Templates/Controllers/Components/Service Providers/Core Theme Files tends to cause a lot of confusion for those new to SquareOne, and it makes onboarding a pretty rough task. 

As a response to this, and in an attempt to streamline our system in order to provide easier onboarding and more conducive workflows for developers working on the project, we've decided to adopt a pattern more similar to the standard MVC or MCC systems you see in systems like Laravel or Symfony. 

This is a pretty big shift, and there are a lot of moving parts! As such, I wrote up a walkthrough that you can find in `docs/concepts/templating.md`. That'll give you an idea of what this is all about. 

## Tests

Does this have tests?

- [ ] Yes
- [ x] No, I need help figuring out how to write the tests (actually we mostly just need to find time to write coverage)


docs/concepts/templating.md
```diff
@@ -0,0 +1,298 @@
...
+```
+
+The `component()` method takes 2 arguments: the Component path, and the arguments to pass through. We want to load the Masthead component, so we pass in the path to that
+component (keeping the `.php` extension in place!). Second, we have a bunch of arguments we need to pass to the Masthead component so it can render things. Here we pass 
+`masthead`. 
+
+But wait - where did that `mastead` value come from? We didn't set it in `init()`...
+
+And here we see our first example of the cascading Component system in play. If you were to do a dump of the `$data` property before rendering, you'd see that it has a 
+`masthead` value, which we can then use as a variable in the Twig markup. Thgt value, as it happens, comes from none other than our Index Controller! Check it out if you
```
nickpelton, line 267: "Thgt value" > "That value"

docs/concepts/templating.md
```diff
@@ -0,0 +1,298 @@
...
+		language_attributes();
+
+		return ob_get_clean();
+	}
+
+	protected function get_body_class() {
+		return implode( ' ', get_body_class() );
+	}
+
+	public function render(): void {
```
msawicki, line 175: The biggest thing that gives me pause about this new architecture is this. Adding twig in php. Did you give any thoughts to doing something like `echo $this-render('my/template/index.twig', $this-data);`. It's a bit closer to other MVC frameworks and very similar to `$this->render_component( 'document/Document.php', $args );`

docs/concepts/templating.md
```diff
@@ -0,0 +1,298 @@
...
+
+```twig
+{{ component( 'header/masthead/Masthead.php', masthead ) }}
+```
+
+The `component()` method takes 2 arguments: the Component path, and the arguments to pass through. We want to load the Masthead component, so we pass in the path to that
+component (keeping the `.php` extension in place!). Second, we have a bunch of arguments we need to pass to the Masthead component so it can render things. Here we pass 
+`masthead`. 
+
+But wait - where did that `mastead` value come from? We didn't set it in `init()`...
```
msawicki, line 264: typo `masthead`

wp-content/plugins/core/src/Router/Routes.php
```diff
@@ -0,0 +1,30 @@
...
+ * Class Routes
+ *
+ * @package Tribe\Project\Router
+ */
+class Routes {
+
+	/**
+	 * @param Router $router
+	 */
+	public function add_routes( Router $router ) {
```
msawicki, line 15: Does this allow passing in dependencies to the Controller? What would that look like? Some Controllers do a bit more "Model' logic that i feel could be moved into those classes.

speerface: Excellent question - and I think that's a very good point. One of the big issues I had with the existing system was the inability to inject dependencies to the Controllers, and I absolutely can see how that would be a good idea here. 

It'll take a bit of refactoring, but since we're already defining the Controllers we want to use in the Definer anyways, it makes sense to me to just update it to allow injection there.

docs/concepts/templating.md
speerface: So we went back and forth on this about 10 times. I think every day when I wake up I have a different opinion. The advantage of having the Twig in the PHP is that it's all self-contained, and there's no juggling of files to get your view up while working with your component data. It basically acts like a cleaner, more-structured version of how template-parts work in most WP themes. 

However, there are some disadvantages, such as Twig autocompleting/typehinting not working in most IDEs, and of course just the sort of weirdness of having markup in your PHP. 

I think I could go either way, and I'm happy to go with the approach that the majority here decides on. It's relatively minor to just pull the Twig out to it's own file once again and call it from within the Component, so I'm perfectly happy doing so if we decide that's the best route!

speerface: Follow-up on this: I'm having Sam and Ryan try out the separate-Twig-file approach as they work on some Blocks to see how it all feels. I'll use that as a good determining factor on whether it makes more sense to separate things out.Stay tuned!

wp-content/plugins/core/src/Router/Routes.php
msawicki: Cool, I haven't totally wrapped my head around the Definers yet.

wp-content/themes/core/components/content-block/Content_Block.php
```diff
@@ -9,45 +9,41 @@
...
+		];
+	}
+
+	public function render(): void {
+		?>
+        <{{ tag }} {{ classes|stringify }} {{ attrs|stringify }}>
+        {{ component( 'text/Text.php', leadin ) }}
+        {{ component( 'text/Text.php', title ) }}
+        {{ component( 'text/Text.php', text ) }}
+        {{ component( 'button/Button.php', action ) }}
```
dpellenwood, line 45: I'm not sure if this was an intentional decision or a demonstration of what is possible. However, forcing a particular component usage within another component will be problematic. For example, we often wrap our [action (button/link) in another component (Text, for example)](https://github.com/moderntribe/square-one/pull/501/files#diff-819c8e311aa7b13860d796dce47980edR102) when the design or style warrants it.

In short, I'm not sure if these "low-level" components can support a specific child-component per usage pattern. 

Additionally, if there is no action data passed into this component, an empty button is still rendered.

wp-content/themes/core/components/container/Container.php
```diff
@@ -8,33 +8,35 @@
...
+			self::TAG               => 'div',
+			self::CLASSES           => [ 'c-container' ],
+			self::ATTRS             => [],
+		];
+	}
+
+	public function render(): void {
+		?>
+		<{{ tag }} {{ classes|stringify }} {{ attrs|stringify }}>
+			{{ component( content_component, content ) }}
```
dpellenwood, line 38: Why better than a forced component usage pattern (see my note on the the Content_Block component below), it feels strange to me to define a "default" component (or really any component at all) at this level. It feels contrary to the idea that the components are dumb templates, rendering whatever data is passed to them without any opinion.

docs/concepts/templating.md
dpellenwood: I _think_ Ryan, Sam and I have landed on keeping separate twig files rather than rendering twig inside PHP. The pain of IDE syntax intellisense alone is enough to say this is probably not worth it.

wp-content/plugins/core/src/Components/Components_Subscriber.php
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Class Components_Subscriber
+ *
+ * @package Tribe\Project\Components
+ */
+class Components_Subscriber extends Abstract_Subscriber {
+
+	public function register(): void {
+		add_filter( 'tribe/libs/twig/funcitons', function ( $fs ) {
```
dpellenwood, line 19: `tribe/libs/twig/funcitons` Typo?

speerface: It's actually a typo in base Gutenpanels, so I had to replicate it here to get it to fire. Once that's fixed in GP we can fix it here, too :)

wp-content/themes/core/components/container/Container.php
speerface: Great question! 

So, in this case, it's because what gets output here needs to be variable by nature, even within a single project. It's a decision that needs to get made at a higher level than this component every time, so we can't have a specific one hardcoded into this Component itself. 

As to the default, I could definitely see an argument being made to have it default to blank, and then maybe just wrap the output in a check for that value being defined. A default helps make it a little less "wait, why is nothing showing?", but I think your point is definitely valid and I wouldn't be against making a move towards that.

wp-content/plugins/core/src/Templates/Components/Component.php
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	/**
+	 * Merge associative arrays of attributes into a string
+	 *
+	 * @param array ...$attributes
+	 *
+	 * @return string
+	 */
+	protected function merge_attrs( array ...$attributes ): string {
```
dpellenwood, line 85: If I understand the change in [this commit](https://github.com/moderntribe/square-one/pull/501/commits/37b8d1605f5932a95b5ab6fb1a6d2dd347fc403d), these two methods (`merge_attrs` & `merge_classes`) are no longer needed, correct?

wp-content/themes/core/components/content-block/Content_Block.php
speerface: Great question! So this should entirely be viewed as simply a starting point. If you needed to change how this is output for your project, you absolutely can/should. If your project is set up in a way where this, for instance, always needs a button wrapped in another component, then you could simply update this to do so. If, on the other hand, your project needs the entire contents of this markup to be a string that you pass in already rendered and just echoed here, you could do that as well. 

I think what it'll come down to is how we think the base/default/most-used pattern should work so that it applies to the most use-cases. I'm not sure I have (or should have) a strong opinion on this, so I'll defer to whatever the FE teams thinks in regards to default setup. 

My main goal with v1 here was just to get everything rendering like it does in the old SquareOne version, and then we can pivot from there as necessary. 

This is exactly the convo I was hoping to have, so thank you for raising these questions!

wp-content/plugins/core/src/Templates/Components/Component.php
speerface: You do, and they are! Great catch.

dpellenwood: Woot! I love the Twig filters for this purpose, btw. It's a really nice enhancement.

wp-content/plugins/core/src/Components/Component.php
```diff
@@ -0,0 +1,184 @@
...
+			$ext = pathinfo( $file, PATHINFO_EXTENSION );
+
+			return $ext === 'twig';
+		} );
+
+		if ( empty( $twigs ) ) {
+			throw new \ErrorException( 'Could not locate a valid Twig template for this component. Please ensure a .twig file exists in the same directory, or that you explicitely define the $twig_file property for this component.' );
+		}
+
+		return sprintf( '%s/%s', $dir, basename( array_shift( $twigs ) ) );
```
dpellenwood, line 93: I noticed while in development, that this method will pull the first twig file which starts with the same string as the component file.

For example, if I have a `Link.php` component with 2 twig files in the same directory: http://p.tri.be/upSuSF
* `Link.php`
* `link.twig`
* `link-with-wrapper.twig`

By default this method is returning `link-with-wrapper.twig` instead of `link.twig`. I think the default template file probably needs to be a direct file name match to the PHP file rather than the first match in the array of twig files.

msawicki:
🍾

PR #503, opened 2020-07-06
Feeture/port other blocks to mvc

faction23:
Ports Interstitital and hero to new arch. I'll leave accordion for you ryan so you get a feel for how it works. Need to get this merged. Thanks!


faction23:
> We're going to go with a twig template yes? Assuming this is just awaiting a tweak from Aaron to ensure they are working properly as I ran into some trouble trying to do so with Media + Text.

Yeep!

PR #516, opened 2020-07-22
disable all gravityform styles - SQONE-308

jbrinley:
## What does this do/fix?

Disables all gravity form styles by filtering the option to enable/disable them. This is to address form styling issues in the block editor. https://moderntribe.atlassian.net/browse/SQONE-308?focusedCommentId=75871

This has the consequence of disabling _all_ of the gform style, and I'm not sure if that's really what we want. We had been dequeuing just `gforms_formsmain_css`. This filter will also dequeue:

- `gforms_reset_css`
- `gforms_datepicker_css`
- `gforms_ready_class_css`
- `gforms_browsers_css`

It looks OK to me, but I don't have the knowledge to know if this is correct.

@jamiepaul, can you let me know if this works?

### DIFF
--- a/wp-content/plugins/core/src/Integrations/Gravity_Forms/Form_Styles.php
+++ b/wp-content/plugins/core/src/Integrations/Gravity_Forms/Form_Styles.php
@@ -22,6 +22,14 @@ class Form_Styles {
 	}
 
 	/**
+	 * @return string
+	 * @filter pre_option_rg_gforms_disable_css
+	 */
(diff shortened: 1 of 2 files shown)


ryanurban:
I think this does make sense. @jamiepaul does GF have an option for whether or not the form is rendered in Gutenberg? Seem to recall reading something about that, if not cool, if so, we may need to do a conditional check on that option before 100% disabling OR just set that option to always be true programmatically.

ryanurban:
Cool, let's do it!

PR #521, opened 2020-07-24
Feeture/sqone 442/panels to blocks

dpellenwood:
## What does this do/fix?
* Renames & moves all the "panels" code in FF to "blocks".
* Removes some old/unused "panel" templates.

[SQONE-442]

## Tests

Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests because...
- [ ] No, I need help figuring out how to write the tests.

[SQONE-442]: https://moderntribe.atlassian.net/browse/SQONE-442

PR #522, opened 2020-07-26
Feeture/sqone 257/quote block

dpellenwood:
## What does this do/fix?

Builds out the FE for the Quote Block. Updates the Quote component to be more flexible.

Based from #521 as this inherits the Panel --> Block renaming.

## QA

Links to relevant issues
- [SQONE-257]

## Tests

Does this have tests?

- [ ] Yes

[SQONE-257]: https://moderntribe.atlassian.net/browse/SQONE-257

### DIFF
--- a/wp-content/themes/core/components/quote/Quote.php
+++ b/wp-content/themes/core/components/quote/Quote.php
@@ -54,20 +54,31 @@ class Quote extends Component {
 	}
 
 	private function setup_quote_text(): array {
+		/**
+		 * Note: HTML5 validation requires that the The block's `<section>` element contain a heading.
+		 * Therefore, we're setting the quote text to an `<h2>`. This means that the quote text cannot be multiple paragraphs.
+		 * However, given the designs for this block, and the notes on the design requesting a max number of lines
+		 * The `<h2>` designation doesn't seem like a limitation.
+		 */
 		return [
-			Text::TEXT    => self::QUOTE_TEXT,
-			Text::CLASSES => 'c-quote__text',
+			Text::TAG     => 'h2',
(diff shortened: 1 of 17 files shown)


wp-content/themes/core/components/quote/Quote.php
```diff
@@ -3,31 +3,111 @@
...
+	}
+
+	public function init() {
+		$this->data[ self::QUOTE ] = $this->setup_quote_text();
+		$this->data[ self::CITE ]  = $this->setup_quote_citation();
+	}
+
+	private function setup_quote_text(): array {
+		/**
+		 * Note: HTML5 validation requires that the The block's `<section>` element contain a heading.
```
jbrinley, line 58: I did not know that `<section>` elements required headings. Where is this requirement coming from? An accessibility standard?

dpellenwood: Yeah, it's a part of the [spec for the `<section>` element](https://www.w3.org/TR/2014/PR-html5-20140916/sections.html#the-section-element):
> The theme of each section should be identified, typically by including a heading (h1-h6 element) as a child of the section element.

The HTML5 validator, and some of the more stringent a11y checkers (Site Improve, Dubbot) will flag it. In practice, my interpretation of the spec & WCAG requirements is that _any_ "sectioning" element should be identifiable. Usually by including a heading tag inside the element, but alternatives include using `aria-label` or `aria-lablledby`.

We even ran into an issue on UCI where articles with their heading tag nested several child elements deep were flagged by Site Improve as not identifiable. We had to add unique-id's to each article's heading tag and then use `aria-labelledby` on the article container to pass the audit.

PR #524, opened 2020-07-29
Feeture/SQONE-437/Stats Block

jamiepaul:
## What does this do/fix?

Adds new Stats block. Includes toolbar options for stacked/inline layouts and content aligned left/center. This block will enable users to display various statistic value + labels in a stylized list next to a content block.

Statistic list items were built with some assumptions:
- Numbers will be used as the value, not text. In the future, this could be enforced by using a number field in the editor. 
- If numbers are too long to fit in the cell, they break to the next line
- Labels can be any length and will wrap lines
- If either number/label wraps to a new line, the height of all elements in the row matches in height and content is aligned to the top of the cell. 

Not included:
- Design specifies that there should be a display option for the stats list to have border enabled/disabled. Borders are added by default. We can add a setting to disable them after we figure out the approach for editor fields.
- Light/dark theme has not been implemented yet as that's something FE needs to holistically plan out.

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/SQONE-437)

Screenshots/video:
- [Link to Screencast Overview](http://p.tri.be/H0xtXS)

## Tests

Does this have tests?

- [ ] Yes


wp-content/themes/core/components/statistic/Statistic.php
```diff
@@ -0,0 +1,51 @@
...
+ * Class Statistic
+ *
+ * @property string   $tag
+ * @property string[] $classes
+ * @property string[] $attrs
+ * @property string[] $container_classes
+ * @property array    $value
+ * @property array    $label
+ */
+class Statistic extends Component {
```
dpellenwood, line 17: Interesting to create a "statistic" component versus just being an element of the stats block. Curious if you foresee using this outside the context of the block or another reason for a separate component?

jamiepaul: Initially I had looked at a few component libraries and saw others included it as its own component. At the time I thought it could potentially be designed to be used in a card header, maybe a spin off of media + text where stat replaces media. About halfway through building it, I started second guessing if it really needed to be separate. Since I was less involved in discussions on what components should be, I'm open to your opinion and @ryanurban's on simplifying to just a part of the Stats block.

dpellenwood:
@jamiepaul Do we still need this PR? Or will #570 handle all of this as well?

jamiepaul:
@dpellenwood nope, we don't need this one as #540 will handle all of this. Going to keep it up until @ryanurban has a chance to see the statistic component comments. I'll be sure to close the PR and delete the branch after that.

wp-content/themes/core/components/statistic/Statistic.php
ryanurban: Let's start with it as a component. Does give us a lot of additional flexibility.

ryanurban:
I believe given response to above this is ready to be closed out @jamiepaul?

jamiepaul:
Closing out as #570 handles the stats block w/ pivoted approach

PR #527, opened 2020-08-03
Adds accordion block

msawicki:
## What does this do/fix?

Adds Accordion Block


wp-content/themes/core/assets/img/admin/blocks/layout-inline.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 134 92"><path d="M69.9 24.3V35h50.5V24.3H69.9zM95.2 31H73.4v-2.4h21.8V31zm17.2 1.3l2.4-4.2 2.4 4.2h-4.8zM69.9 56.9v10.7h50.5V56.9H69.9zm25.3 6.2H73.4v-2.4h21.8v2.4zm19.6 1.8l-2.4-4.2h4.9l-2.5 4.2zM69.9 68.9v10.7h50.5V68.9H69.9zm25.3 6.2H73.4v-2.4h21.8v2.4zm19.6 1.8l-2.4-4.2h4.9l-2.5 4.2zM119.3 36v18.7H70.9V36h48.4m1-1H69.9v20.7h50.4V35z"/><path d="M73.5 41.5H111v1.6H73.5zM73.5 38.4H111V40H73.5zM73.5 44.7h40.6v1.6H73.5zM73.5 47.8h37.3v1.6H73.5zM15.4 29.4h39v2h-39zM15.4 25.4h43v2h-43zM15.4 33.4h45v2h-45zM15.4 41.5h37.4v2H15.4zM15.4 37.5h43v2h-43zM15.4 53.5h37.5v2H15.4zM15.4 49.5h43v2h-43zM15.4 57.5h45v2h-45zM40.3 12.3h55.5v4.8H40.3z"/></svg>
```
dpellenwood, line 1: Curious if you're creating these SVGs, or where you're getting them from?

wp-content/plugins/core/src/Blocks/Types/Accordion/Accordion.php
```diff
@@ -3,78 +3,95 @@
...
-			->add_data_source( 'className-c-block', self::LAYOUT )
-			->add_toolbar_section( $this->layout_toolbar() )
-			->add_content_section( $this->content_area() )
-			->add_content_section( $this->accordions_area() )
-			->build();
+	public function add_block() {
+		$this->set_block( new Block( self::NAME, [
+			'title'       => __( 'Accordion', 'tribe' ),
+			'description' => __( 'The Accordion block', 'tribe' ),
+			'icon'        => '<svg width="24" height="18" fill="none" xmlns="http://www.w3.org/2000/svg"><path fill="#000" d="M0 0h23.7v2.7H0zM0 12h23.7v2.7H0zM0 15.3h23.7V18H0z"/><path fill="#fff" stroke="#000" d="M1.7 3.8h20v6.8h-20z"/></svg>',
```
dpellenwood, line 28: Curious where you're getting this icon?

wp-content/plugins/core/src/Blocks/Types/Accordion/Accordion.php
```diff
@@ -3,78 +3,95 @@
...
-	public const DESCRIPTION = 'description';
-	public const ACCORDION   = 'accordion';
+	public const TITLE             = 'title';
+	public const DESCRIPTION       = 'description';
+	public const ACCORDION         = 'accordion';
+	public const ACCORDION_HEADER  = 'accordion_header';
+	public const ACCORDION_CONTENT = 'accordion_content';
 
 	public const LAYOUT         = 'layout';
 	public const LAYOUT_INLINE  = 'layout-inline';
```
dpellenwood, line 21: Let's remove the prefix from these values `layout-`. This was a limitation of the GutenPanels options implementation.

wp-content/plugins/core/src/Blocks/Types/Accordion/Accordion.php
```diff
@@ -3,78 +3,95 @@
...
+class Accordion extends Block_Config {
+	public const NAME = 'accordion';
 
-	public const TITLE       = 'title';
-	public const DESCRIPTION = 'description';
-	public const ACCORDION   = 'accordion';
+	public const TITLE             = 'title';
+	public const DESCRIPTION       = 'description';
+	public const ACCORDION         = 'accordion';
+	public const ACCORDION_HEADER  = 'accordion_header';
```
dpellenwood, line 17: I wonder if we should name these `row_header` & `row_content` or something similar? I feel like `accordion_header` etc is a bit confusing.

wp-content/themes/core/assets/img/admin/blocks/layout-inline.svg
msawicki: They are the icons we used in panels. I'm just porting them over for now as placeholder.

wp-content/plugins/core/src/Blocks/Types/Accordion/Accordion.php
msawicki: I made this one.

wp-content/plugins/core/src/Blocks/Types/Accordion/Accordion.php
msawicki: One thing to keep in mind here is the value must relate to the image. So if we have `left` or `center` then we probably need to specific the path for the images like `/assets/img/admin/blocks/accordion` so they don't conflict with a `left` or `center` specific to another panel.

dpellenwood: Ah, right. I wasn't thinking that thru. Would be nice to be able to specify an image per option similar to the swatch select's color-per-option. Or, maybe, doing a per-block folder in assets isn't a bad idea anyway. It makes the images directory a bit quicker to grep.

For icons/images that aren't specific, they could remain in the root, but add a block-specific directory for any specific options. Thoughts?

PR #528, opened 2020-08-03
Adds Hero block

msawicki:
## What does this do/fix?

Adds Hero Block


wp-content/plugins/core/src/Blocks/Types/Hero/Hero.php
```diff
@@ -3,106 +3,85 @@
...
 
 	public const IMAGE         = 'image';
 	public const LEAD_IN       = 'leadin';
 	public const TITLE         = 'title';
 	public const DESCRIPTION   = 'description';
 	public const CTA           = 'cta';
 	public const LAYOUT        = 'layout';
-	public const LAYOUT_LEFT   = 'layout-left';
-	public const LAYOUT_CENTER = 'layout-center';
+	public const LAYOUT_LEFT   = 'hero-layout-left';
```
dpellenwood, line 20: Can we reduce these to just `left` & `center`?

PR #534, opened 2020-08-04
SQONE-458 Register ACF Media + Text Block

dpellenwood:
## What does this do/fix?

Builds the Media + Text block.

@msawicki I'm having some trouble with the conditional logic for the image versus video field. Can you peak at that?


wp-content/plugins/core/src/Blocks/Types/Media_Text/Media_Text.php
```diff
@@ -3,77 +3,150 @@
...
+					self::IMAGE,
+				],
+			] )
+		)->add_field( new Field( self::NAME . '_' . self::IMAGE, [
+				'label'             => __( 'Image', 'tribe' ),
+				'name'              => self::IMAGE,
+				'type'              => 'image',
+				'preview_size'      => 'medium',
+				'instructions'      => __( 'Recommended image size by layout:<br>Center: 1920px wide with a 16:9 aspect ratio.<br>Left/Right: 1536px wide with a 4:3 aspect ratio.', 'tribe' ),
+				'conditional_logic' => [
```
dpellenwood, line 83: This doesn't appear to work. Thoughts? Same with the Video field below.

msawicki: Fields get prefixed with their type in the ACF_Configuration class so you need to prepend `field_`
`'field_' . self::NAME . '_' . self::MEDIA_TYPE,`
`'field_' . self::NAME . '_' . self::MEDIA_TYPE`

wp-content/themes/core/assets/img/admin/blocks/media-text/full.svg
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 21.1.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
```
msawicki, line 2: I know these are probably just placeholders but should we run these through some kind of minifier? SVGOMG or something?

dpellenwood: These are strait from our existing panel settings. Very FPO, so I didn't want to worry about optimizing them. I'm putting together a list of the stuff we need from design today yet.

wp-content/plugins/core/src/Blocks/Types/Media_Text/Media_Text.php
dpellenwood: Purrfect. thank you!

PR #535, opened 2020-08-05
unify on a consistent factory method

jbrinley:
## What does this do/fix?

- Adds an abstract parent class for all template controllers
- Removes the Factory_Method trait
- Passes all factories to the DI container using the make() method

Our templates will still instantiate the controller the same way:

```
$controller = \Tribe\Project\Templates\Components\button\Controller::factory( $args );
```

Instead of using `$container()->get()`, the factory will use `$container()->make()`. The differences between the two:

1. `make()` will create a new instance each time it is called
2. `make()` allows us to pass arguments to the constructor

But it still supports dependency injection, so we can have a combination of autowired DI and explicit args. Something like:

```
$controller = \Tribe\Project\Templates\Components\something\Controller::factory( $args );
```

And in `something\Controller.php`:

```
class Something extends Abstract_Controller {
  public function __construct( Social_Settings $social, Theme_Build_Parser $parser, array $args ) {}
}
```

Later we can discuss replacing `$args` with named arguments.


jbrinley:
Sorry, this PR blew up a little bit. The first commit is the only relevant one. The rest is to get the test suite passing.

PR #536, opened 2020-08-05
begin the process of purging our aborted attempts at templating

jbrinley:
## What does this do/fix?

Begins purging all the detritus from our prior attempts at templating. This isn't done, but it's a start.


jbrinley:
I like the router thing, too, but:
1. We don't need it right now
2. It belongs in tribe-libs.

PR #537, opened 2020-08-05
Adds Tabs block

msawicki:
## What does this do/fix?

Adds Tab block


wp-content/plugins/core/src/Blocks/Types/Tabs/Tabs.php
```diff
@@ -0,0 +1,106 @@
...
+
+		return $group;
+	}
+
+	/**
+	 * Register Settings for Block
+	 */
+	public function add_settings() {
+		$this->add_setting( new Field( self::NAME . '_' . self::LAYOUT, [
+			'type'            => 'image_select',
```
dpellenwood, line 89: `name` attribute. (If it isn't already added in the merge.)

PR #538, opened 2020-08-05
Save markup in output buffer and actually return it once cached

msawicki:
## What does this do/fix?

Fixes the Oembed filter

@jbrinley question on this. Ryan and I were talking and we can see instances where a "dumb" component may need to be generated in a higher Controller and using `get_template_part` without the output buffer won't work. Do you see any value in having a `tribe_get_template_part` wrapper with an optional `$echo` arg in case this becomes a thing? Obviously we don't want a ton of output sitting in memory but interested on your thoughts here.


ryanurban:
What is left here @msawicki? Need a final CR from Jonathan?

PR #539, opened 2020-08-06
Feeture/finished component class usage

ryanurban:
This PR updates components, where appropriate, to allow for the setting of core required classes while allowing for the ability to add additional classes if needed. We have two types of components:

- Components such as link or button where when used, any base classes _should_ be wiped out. They should not and don't need a core set of always present classes and instead classes should be added for each use to allow for the ultimate flexibility for composing FE.
- Components such as video or card where when used, there is a base set of core classes _required_ to accomplish the base component UI. In this case, we only need the ability occasionally to append additional classes.

@dpellenwood this is the resulting PR based on our convo around this. Note that Jonathan, Mike, and I determined that for now, we probably don't need a markup utility.

### DIFF
--- a/wp-content/themes/core/components/image/Controller.php
+++ b/wp-content/themes/core/components/image/Controller.php
@@ -82,7 +82,7 @@ class Controller extends Abstract_Controller {
 		// append an html string inside the wrapper. Useful for adding a `<figcaption>` or other markup after the image
 		$this->html = $args['html'] ?? '';
 		// pass classes for image tag. if lazyload is true class "lazyload" is auto added
-		$this->img_classes = array_merge( [ 'c-image__image' ], (array) $args['img_classes'] ?? [] );
+		$this->img_classes = array_merge( [ 'c-image__image' ], (array) ( $args['img_classes'] ?? [] ) );
 		// additional image attributes
 		$this->img_attrs = (array) ( $args['img_attrs'] ?? [] );
 		// pass specific image alternate text. if not included, will default to image title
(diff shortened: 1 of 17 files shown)


wp-content/themes/core/components/image/Controller.php
```diff
@@ -117,7 +117,7 @@ public function __construct( array $args = [] ) {
 		$this->use_srcset = $args['use_srcset'] ?? true;
 		$this->wrapper_attrs = (array) ( $args['wrapper_attrs'] ?? [] );
 		// pass classes for figure wrapper. If as_bg is set true gets auto class of "lazyload"
-		$this->wrapper_classes = (array) ( $args['classes'] ?? [ 'c-image' ] );
+		$this->wrapper_classes = array_merge( [ 'c-image' ], (array) ( $args['classes'] ?? [] ) );
```
dpellenwood, line 120: Is there styles associated with `c-image`? This would be once case where I don't want to have to override any default styles.

wp-content/themes/core/twig-components/blocks/wysiwyg/wysiwyg.twig
```diff
@@ -1,13 +1,13 @@
 {# Block: WYSIWYG #}
 
 <div class="s-content">
-	<div class="g-row g-row--center g-row--col-3--min-full"
-			 data-depth="0"
-			 data-name="columns"
-			 data-livetext>
+	<div data-depth="0"
```
dpellenwood, line 4: we can drop these data attributes, yes?

ryanurban: Sure can and they will be as they are translated over from this now legacy directory back to components in removing twig, just hadn't gotten to it yet ; )

wp-content/themes/core/components/image/Controller.php
ryanurban: Not totally sure yet. but given it has so many options and things going on, at least to start, I don't think it hurts to have a commonly available css hook available to do all the things with. We can def revise if we find in project use it goes the way of the painful button 🙃

PR #540, opened 2020-08-06
Block Controller Architecture. 

msawicki:
## What does this do/fix?

Adds the beginnings for the block architecture. Couples with https://github.com/moderntribe/tribe-libs/pull/60

This: 

- Creates a 'block' directory in the theme to handle the rendering of the blocks. 
- Keeps the actual component in the components directory so it doesn't have to be used with just a block
- Adds a "Model" to handle all the data collection for the block, which is then passed to `get_template_part()` in the block file. 
- Adds a "Base_Model" which takes the `$block` array from ACF's default rendering and sets up some specific properties and has a helper method to get data values for a block.
- This fleshes out the Hero block, which is missing a "Content Block" component at the moment so I'm var_dumping.

Overall, the main idea is that block components can be reusable throughout a site and should probably not include the block specific modeling.


wp-content/plugins/core/src/Blocks/Types/Hero/Model.php
```diff
@@ -0,0 +1,134 @@
+<?php
```
ryanurban, line 1: Not sure I follow the purpose for this model or at least in the core plugin? This seems more hero block controller'esque based on what we've been doing thus far, namely with all the setting of classes & markup.

dpellenwood: Yeah. I thought the same when Mike first showed me this arch model today. However, I was brought around to this idea b/c it allows our block "components" to stay data-source agnostic. Meaning, unlike a template component or dumb component, "blocks" need to source their data from somewhere. And we want to keep the block component agnostic of that.

It allows the blocks to be used as a "true" component that doesn't care about the data source. In this way,  the hero block _component_ could power both a page's subheader via meta-fields on the post type _and_ a hero block within the page's content.

ryanurban: Ok, that's helpful, however, that also sounds like what we were moving away from in this pivot. We still have that capability for our non route / content "component"s and can keep it in the theme, which reduces cognitive overhead. For the sake of consistency, I'd vote this goes in the theme like we've done with all the other templates/components now. My two cents. I'll let @jbrinley and @nickpelton weigh in and also vote. Regardless, thanks for knocking out the first pass, Mike ❤️

wp-content/plugins/core/src/Blocks/Types/Hero/Model.php
```diff
@@ -0,0 +1,134 @@
...
+				Image_Sizes::CORE_FULL,
+				Image_Sizes::CORE_MOBILE,
+			],
+		];
+	}
+
+	/**
+	 * @return array
+	 */
+	private function get_content(): array {
```
jbrinley, line 68: I'm a little uncomfortable with all of the `get_*` method names, because they are not getting the thing, _per se_, but the arguments that will be used to build the thing. E.g., `get_content_args()` seems more accurate than `get_content()`, and properly distinguishes its behavior from `get_layout()`, which is returning the actual layout value, not args to build a layout.

wp-content/plugins/core/src/Blocks/Types/Hero/Model.php
```diff
@@ -0,0 +1,134 @@
...
+			'layout'  => $this->get_layout(),
+		];
+	}
+
+	/**
+	 * @return array
+	 */
+	private function get_leadin(): array {
+		return [
+			'tag'     => 'p',
```
jbrinley, line 84: This feels suspiciously like we're writing HTML in our models. I think it merits a discussion later this morning.

msawicki:
Closing in favor of Option 2

PR #541, opened 2020-08-06
Feeture/detwig quote

nickpelton:
## What does this do/fix?

This removes twig from the Quote Component.

## QA

NA

## Tests

Does this have tests?

- [ ] Yes
- [x ] No, this doesn't need tests because...

Too much flux to test yet.


nickpelton:
Note, the image component is in flux, so this is incomplete till @ryanurban finishes it.

nickpelton:
Blargh, phpcs is busted on my local. FIxing.

wp-content/themes/core/components/quote/Controller.php
```diff
@@ -0,0 +1,94 @@
...
+		// Blockquote properties
+		$this->text_tag     = (string) ( $args['text_tag'] ??'h2' );
+		$this->text_text    = (string) ( $args[''] ?? 'Some quote test for you!' );
+		$this->text_classes = (array) ( $args[''] ?? [ 'c-quote__text', 'h4' ] );
+		$this->text_attrs   = (array) ( $args[''] ?? [] );
+
+		// Cite properties
+		$this->cite_name    = (string) ( $args['cite_name'] ?? 'Cite Name' );
+		$this->cite_title   = (string) ( $args['cite_title'] ?? 'Cite Title' );
+		$this->cite_image   = (string) ( $args['cite_image'] ?? get_stylesheet_directory_uri() . '/components/quote/assets/img/default.jpg' );
```
dpellenwood, line 55: Forgive me if I'm reading the conditional wrong, but won't this mean that the cite image will never be empty? I think there will be times that we don't want to use an image with the cite.

wp-content/themes/core/components/quote/Controller.php
```diff
@@ -0,0 +1,94 @@
...
+
+	public function __construct( array $args = [] ) {
+
+		// Component properties
+		$this->classes =  (array) ( $args['classes'] ?? [ 'c-quote' ] );
+		$this->attrs   = (array) ( $args['attrs'] ?? [] );
+
+		// Blockquote properties
+		$this->text_tag     = (string) ( $args['text_tag'] ??'h2' );
+		$this->text_text    = (string) ( $args[''] ?? 'Some quote test for you!' );
```
dpellenwood, line 48: I don't think we should set a default for this value. Or, if we do, then it should def be translatable.  Here and below.

PR #542, opened 2020-08-07
Block Controllers - Option 2

msawicki:
## What does this do/fix?

Here is another option that pushed more component building responsibility to the theme Controller and leaves our model to return purely data.


wp-content/plugins/core/src/Blocks/Types/Hero/Model.php
```diff
@@ -0,0 +1,55 @@
...
+	private function get_content(): array {
+		return [
+			self::LEAD_IN => $this->get( Hero::LEAD_IN, '' ),
+			self::TITLE   => $this->get( Hero::TITLE, '' ),
+			self::TEXT    => $this->get( Hero::DESCRIPTION, '' ),
+			self::CTA     => wp_parse_args( $this->get( Hero::CTA, [] ), [
+				'title'  => '',
+				'url'    => '',
+				'target' => '',
+			] ),
```
ryanurban, line 51: With our new container component + defer_get_template method, should we keep this here or instead just pass `Hero::CTA` on itself to the Hero block controller in the theme?

wp-content/themes/core/components/blocks/hero/Controller.php
```diff
@@ -0,0 +1,167 @@
+<?php
```
ryanurban, line 1: Totally cool with you merging this Mike and then cleaning it up (given where we've landed as of today in our approach) post merge since we still need to do the blocks as the final thing.

PR #545, opened 2020-08-07
Accordion untwigify

msawicki:
## What does this do/fix?

Moves over Accordion component. This is my first attempt at moving this stuff over. Would love feedback.


wp-content/themes/core/components/accordion/accordion.php
```diff
@@ -0,0 +1,55 @@
...
+// phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UndefinedVariable
+$controller = \Tribe\Project\Templates\Components\accordion\Controller::factory( $args );
+?>
+<div
+	<?php echo $controller->container_classes(); ?>
+		role="tablist"
+		aria-multiselectable="true"
+	<?php echo $controller->container_attrs(); ?>
+>
+	<?php foreach ( $controller->rows as $key => $row ) { ?>
```
dpellenwood, line 11: This is super minor, but I prefer the alternate loop syntax where ever we are using php open & close tags within the loop: `<?php foreach ( $controller->rows as $key => $row ): ?> ... <?php endforeach; ?>`. Curious to hear other's thoughts on this.

wp-content/themes/core/components/accordion/accordion.php
```diff
@@ -0,0 +1,55 @@
+<?php
+// phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UndefinedVariable
+$controller = \Tribe\Project\Templates\Components\accordion\Controller::factory( $args );
+?>
+<div
```
dpellenwood, line 5: Oof this is difficult to read/grok quickly. I winder if we should just pass in the additional attributes from the controller instead of mixing the static & dynamic attributes here?

wp-content/themes/core/components/accordion/accordion.php
```diff
@@ -0,0 +1,55 @@
...
+	<?php echo $controller->container_classes(); ?>
+		role="tablist"
+		aria-multiselectable="true"
+	<?php echo $controller->container_attrs(); ?>
+>
+	<?php foreach ( $controller->rows as $key => $row ) { ?>
+	<article <?php echo $controller->row_classes(); ?>>
+		<<?php echo esc_attr( $controller->row_header_tag ); ?>>
+		<button
```
dpellenwood, line 14: Similar to the above comment, I wonder if we setup all these attributes in the controller instead of mixing them here? Alternatively, can the dynamic (classes output) to the first or last attribute and the white space aligned? (This applies below as well.)

wp-content/themes/core/components/accordion/accordion.php
```diff
@@ -0,0 +1,55 @@
...
+				aria-controls="<?php echo esc_attr( $row[ 'content_id' ] ); ?>"
+				aria-expanded="false"
+				aria-selected="false"
+			<?php echo $controller->row_header_classes(); ?>
+				id="<?php echo esc_attr( $row[ 'header_id' ] ); ?>"
+				data-index="<?php echo absint( $key ); ?>"
```
dpellenwood, line 20: I think we can drop the data-index here & similarly, the `data-name`, `data-index`, and `data-livetext` attributes throughout this component as those are specifically for panel builder.

wp-content/themes/core/components/accordion/accordion.php
jbrinley: I disagree. Largely because in the IDE it's easy to put a cursor on `}` and have it highlight the matching `{`.

ryanurban: I also prefer `foreach` 😂

wp-content/themes/core/components/accordion/Controller.php
```diff
@@ -0,0 +1,142 @@
...
+			'row_header_tag'                => 'h3',
+			'row_classes'                   => [ 'c-accordion__row' ],
+			'row_header_classes'            => [ 'c-accordion__header', 'h5' ],
+			'row_header_container_classes'  => [ 'c-accordion__header-container' ],
+			'row_content_classes'           => [ 'c-accordion__content' ],
+			'row_content_container_classes' => [ 'c-accordion__content-container', 't-sink', 's-sink' ],
+			'row_header_name'               => 'title',
+			'row_content_name'              => 'row_content',
+		];
+	}
```
ryanurban, line 113: These defaults can be setup as part of the constructor above (scope out any of the other done components).

wp-content/themes/core/components/accordion/accordion.php
ryanurban: Agreed and they should be set as required attributes

wp-content/themes/core/components/accordion/accordion.php
faction23: I prefer `{  }` as well, 1 char, loop block highlighting, and matches js/other langs

wp-content/themes/core/components/accordion/accordion.php
ryanurban: Yep!

wp-content/themes/core/components/accordion/Controller.php
ryanurban: Fair enough, it is cleaner more clear. I think the main reason I've liked the other is the reduction in code in the controller. I could go either way. @nickpelton @dpellenwood @jbrinley?

dpellenwood: I'm also finding this true as I've worked on the Tabs component. I agree with you, Mike. That being said, we need to keep the ability to both allow defaults and required classes/attributes for each component. In this case, the majority of the values here should be required, not just defaults as the component's JS requires the attrs to function properly and the component's styles rely on these classes.

dpellenwood: For context, this is the tabs component with the current arch (no defaults method) showing the required and default properties all inlined. I think this is really, really hard to read:
```
	public function __construct( array $args = [] ) {
		$this->tabs                = (array) $args['tabs'] ?? [];
		$this->container_classes   = array_merge( ['c-tabs'], (array) $args['container_classes'] ?? [] );
		$this->container_attrs     = array_merge( [ 'data-js' => 'c-tabs' ], (array) $args['container_attrs'] ?? [] );
		$this->tab_list_classes    = array_merge( ['c-tab__list'], (array) $args['tab_list_classes'] ?? [] );
		$this->tab_list_attrs      = (array) $args['tab_list_attrs'] ?? [];
		$this->tab_button_classes  = array_merge( ['c-tab__button'], (array) $args['tab_button_classes'] ?? [] );
		$this->tab_button_attrs    = (array) $args['tab_button_attrs'] ?? [];
		$this->tab_content_classes = array_merge( ['c-tab__content'], (array) $args['tab_button_classes'] ?? ['s-sink', 't-sink'] );
		$this->tab_content_attrs   = array_merge( [ 'role' => 'tabpanel', 'tabindex' => 0 ], (array) $args['tab_content_attrs'] ?? [] );
		$this->tab_list_buttons    = $this->get_tablist_buttons();
	}
```

nickpelton: I also like this style better. The heavy use of the `??` was mostly for speed before we go back and refactor common problems back into the `Abstract_Controller`. After doing it enough times, I think this approach makes sense as long as we can still have defaults, override defaults, and some required arguments.

wp-content/themes/core/components/accordion/accordion.php
```diff
@@ -0,0 +1,55 @@
...
+				>
+					<?php echo $row[ 'header_text' ]; ?>
+				</span>
+		</button>
+		</<?php echo esc_attr( $controller->row_header_tag ); ?>>
+		<div
+				hidden
+				aria-hidden="true"
+				aria-labelledby="<?php echo esc_attr( $row[ 'header_id' ] ); ?>"
+			<?php echo $controller->row_content_classes(); ?>
```
nickpelton, line 38: General Component Nit. I would like to shorten the `$controller` name for template readability. Be nice if it was just `$c->foo`.

wp-content/themes/core/components/accordion/Controller.php
dpellenwood: Proposal: We setup the defaults this way and inject the required properties within the markup util methods (below)?

jbrinley: Plus one to this style. What about having two conventions?

1. A `defaults()` method that returns an array of defaults that can be overwritten with the args.
2. A `merge()` method that returns an array of arrays that will be merged with the args.

dpellenwood: Yep. This is better, Jonathan. I was just getting to this as I was adjusting the tabs component. My idea of merging inside the markup utils wasn't ideal either.

PR #546, opened 2020-08-08
Container component, wrapper method

jbrinley:
## What does this do/fix?

Refactors the container component to the new architecture. This adds a helper function, `defer_template_part()`, which allows us to pass a call to render a template part as an object to another template. Here's what it would look like to wrap a button in a container:

```
get_template_part( 'components/container/container', null, [
	'classes' => [ 'test-button' ],
	'attrs'   => [ 'data-test' => 'value' ],
	'content' => defer_template_part( 'components/button/button', null, [
		'content'    => 'Click Me!',
		'aria-label' => 'This is a button',
	] ),
] );
```

How else might this be useful? Let's say we have an Image + Text block that needs to use the "content_block" component to render part of its markup.

Our controller for Image + Text might do something like:

```
public function content(): string {
	return tribe_template_part( 'components/content_block/content_block', null, [
		'title' => $this->get_title(),
		'link'  => $this->get_link(),
	] );
}

private function get_title(): Deferred_Component {
	return defer_template_part( 'components/text/text', null, [
		'text' => $this->title,
	] );
}

private function get_link(): Deferred_Component {
	return defer_template_part( 'components/link/link', null, [
		'url'    => $this->url,
		'target' => $this->target,
	] );
}
```

And then our `content_block\Controller` might do:

```
public function title() {
	if ( ! isset( $this->title['tag'] ) ) {
		$this->title['tag'] = 'h3';
	}
	$this->title['classes'][] = 'content-block__title';
	return $this->title->render(); // or just: (string) $this->title;
}
public function link() {
	$this->link['classes'][] = 'content-block__link';
	return $this->link->render(); // or just: (string) $this->link;
}
```


wp-content/plugins/core/functions/templates.php
```diff
@@ -14,3 +17,16 @@ function tribe_template_part( $slug, $name = null, $args = [] ) {
...
+
+if ( ! function_exists( 'defer_template_part' ) ) {
+	/**
+	 * @param string      $slug
+	 * @param string|null $name
+	 * @param array       $args
+	 *
+	 * @return Deferred_Component
+	 */
+	function defer_template_part( $slug, $name = null, $args = [] ) {
```
jbrinley, line 29: Ryan and I had a discussion about naming this function, but there might be better options. Suggestions?

dpellenwood: Agreed, I'm not sure the name aligns with the use quite right. Maybe something like `nest_template_part()`?

ryanurban: I prefer `defer` or maybe stash` to `nest` as this can be utilized in other contexts than just the container component.

dpellenwood: Hmm... I'm not sure I follow that. In what context would the output from this not be _nested_ within another component?

wp-content/themes/core/components/container/Controller.php
```diff
@@ -0,0 +1,50 @@
...
+	/**
+	 * @var string[]
+	 */
+	private $attrs;
+	/**
+	 * @var string|Deferred_Component
+	 */
+	private $content;
+
+	public function __construct( array $args = [] ) {
```
nickpelton, line 28: This makes me very happy after doing this on every component controller. @ryanurban This is where we should align around common methods for managing classes and other consistent parts of a component.

wp-content/plugins/core/functions/templates.php
nickpelton: I'm fine with defer.

wp-content/plugins/core/src/Templates/Components/Deferred_Component.php
```diff
@@ -0,0 +1,41 @@
...
+	/**
+	 * @var string
+	 */
+	private $path;
+	/**
+	 * @var string|null
+	 */
+	private $name;
+
+	public function __construct( string $path, string $name = null, array $args = [] ) {
```
nickpelton, line 16: This will also allow us to reduce quite a bit of component template noise with this container. +1

wp-content/themes/core/components/container/container.php
```diff
@@ -0,0 +1,13 @@
...
+declare( strict_types=1 );
+
+/**
+ * @var array $args Arguments passed to the template
+ */
+// phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UndefinedVariable
+$controller = \Tribe\Project\Templates\Components\container\Controller::factory( $args );
+
+?>
+<<?php echo $controller->tag(); ?> <?php echo $controller->classes(); ?> <?php echo $controller->attributes(); ?>>
```
nickpelton, line 11: Formatting. I've noticed we often break this calls to multiple lines for readability. @ryanurban can confirm.

ryanurban: Yeah, I can do a pass on this later, post merge, it's a react thing : )

PR #547, opened 2020-08-11
Feeture/pagination untwig

msawicki:
## What does this do/fix?

Adds a single pagination "component"

### DIFF
--- a/wp-content/themes/core/components/pagination/single/single.php
+++ b/wp-content/themes/core/components/pagination/single/single.php
@@ -1,0 +1,35 @@
+<?php
+// phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UndefinedVariable
+$c = \Tribe\Project\Templates\Components\pagination\single\Controller::factory( $args );
+
+?>
+<?php if ( ! empty( $c->previous_post ) || ! empty( $c->next_post ) ) { ?>
+	<nav <?php echo $c->wrapper_classes(); ?><?php echo $c->wrapper_attrs(); ?> >
+
+		<h3 <?php echo $c->header_attrs(); ?><?php echo $c->header_classes(); ?>>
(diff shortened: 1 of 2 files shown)


wp-content/themes/core/components/pagination/single/single.php
```diff
@@ -0,0 +1,35 @@
+<?php
+// phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UndefinedVariable
+$c = \Tribe\Project\Templates\Components\pagination\single\Controller::factory( $args );
+
+?>
+<?php if ( ! empty( $c->previous_post ) || ! empty( $c->next_post ) ) { ?>
+	<nav <?php echo $c->wrapper_classes(); ?><?php echo $c->wrapper_attrs(); ?> >
+
+		<h3 <?php echo $c->header_attrs(); ?><?php echo $c->header_classes(); ?>>
```
msawicki, line 9: I kind of made this component up, do we want any tag customization on the header or the wrapper?

wp-content/themes/core/components/pagination/single/single.php
```diff
@@ -0,0 +1,35 @@
...
+			<?php esc_html_e( 'Post Pagination', 'tribe' ); ?>
+		</h3>
+
+		<ol <?php echo $c->container_classes(); ?><?php echo $c->container_attrs(); ?>>
```
msawicki, line 13: Will this always be ol? Do we need an option where this is an ul?

ryanurban: Nope, will always be ol

wp-content/themes/core/components/pagination/single/single.php
ryanurban: Nope, and I'll actually nix the `h3` in favor of adding a `aria-label` and putting the static header text as the value

PR #549, opened 2020-08-11
Feeture/content block untwig

ryanurban:
This PR implements our newly agreed upon approach to the done component controllers and implements a first pass for the content block component.


wp-content/themes/core/components/container/Controller.php
```diff
@@ -26,10 +26,31 @@ class Controller extends Abstract_Controller {
...
+			'tag'     => '',
+			'classes' => [],
+			'attrs'   => [],
+			'content' => '',
+		];
+	}
+
+	protected function required(): array {
+		return [
+			'tag' => 'div',
```
msawicki, line 52: If the tag is required to be a div, shouldn't we just pull our this option and make it a div in the template?

wp-content/themes/core/components/content_block/Controller.php
```diff
@@ -0,0 +1,133 @@
...
+			'leadin'  => null,
+			'title'   => null,
+			'content' => null,
+			'cta'     => null,
+		];
+	}
+
+	protected function required(): array {
+		return [
+			'tag'     => 'div',
```
msawicki, line 87: same comment here. If div is required, let's make it static and remove the option.

wp-content/themes/core/components/image/Controller.php
```diff
@@ -6,120 +6,216 @@
...
+			$args[$key] = array_merge( $args[$key], $value );
+		}
+
 		// the WordPress attachment to use - takes precedence over IMG_URL
-		$this->attachment = $args['attachment'] ?? null;
+		$this->attachment = $args['attachment'];
+		$this->attrs = (array) $args['attrs'];
+		// pass classes for figure wrapper. If as_bg is set true gets auto class of "lazyload"
+		$this->classes = (array) $args['classes'];
+		// html tag for the wrapper/background image container
```
msawicki, line 131: All these comments probably belong on defaults or in the properties themselves. What do you think?

wp-content/themes/core/components/routes/unsupported_browser/unsupported_browser.php
```diff
@@ -50,10 +49,10 @@ class="browser-list__item-anchor"
 					   rel="external noopener"
 					   target="_blank">
 						<span class="browser-list__item-image">
-						  <img src="<?php echo $controller->legacy_browser_icon_chrome ?>"
-							alt="<?php echo esc_attr( __( 'Chrome browser logo', 'tribe' ) ) ?>" />
+						  <img src="<?php echo $c->legacy_browser_icon_chrome; ?>"
```
msawicki, line 52: Do we need `esc_url()` here?

wp-content/themes/core/components/routes/unsupported_browser/unsupported_browser.php
```diff
@@ -39,8 +38,8 @@ class="site-logo site-logo--header"
 		<div class="l-container">
 
 			<div class="site-content__content">
-				<h2><?php echo $controller->legacy_browser_title ?></h2>
-				<p><?php echo $controller->legacy_browser_content ?></p>
+				<h2><?php echo $c->legacy_browser_title; ?></h2>
```
msawicki, line 41: `esc_html()`?

wp-content/themes/core/components/content_block/Controller.php
ryanurban: Hah, that's just my bad, good catch. Set it to be the default vs. required.

wp-content/themes/core/components/container/Controller.php
ryanurban: Same here, good catch.

wp-content/themes/core/components/image/Controller.php
ryanurban: Good call, will put on the properties I think

wp-content/themes/core/components/routes/unsupported_browser/unsupported_browser.php
ryanurban: I don't think so, it does get escaped in the controller.

wp-content/themes/core/components/routes/unsupported_browser/unsupported_browser.php
ryanurban: Fixed

ryanurban:
@msawicki updated based on your notes. Want to do another check or gtg?

PR #552, opened 2020-08-12
Implement subheader page title and update text component

ryanurban:
This implements the `page_title` method as a trait that is now globally available and makes use of it along with a proper page title in the subheader component. I also updated the text component based on something I missed as part of the greater done component clean up pass.

Dave, I could see us adding some additional args and even variations to subheader in the future, but I think can wait until we start using and finish building out the base theme.

### DIFF
--- a/wp-content/themes/core/components/header/subheader/Controller.php
+++ b/wp-content/themes/core/components/header/subheader/Controller.php
@@ -4,12 +4,20 @@ declare( strict_types=1 );
 namespace Tribe\Project\Templates\Components\header\subheader;
 
 use Tribe\Project\Templates\Components\Abstract_Controller;
+use Tribe\Project\Templates\Components\Traits\Page_Title;
 
 class Controller extends Abstract_Controller {
+	use Page_Title;
 
-	public function title(): string {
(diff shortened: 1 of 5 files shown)


wp-content/themes/core/components/header/subheader/Controller.php
```diff
@@ -4,12 +4,20 @@
...
+use Tribe\Project\Templates\Components\Traits\Page_Title;
 
 class Controller extends Abstract_Controller {
+	use Page_Title;
 
-	public function title(): string {
-		// TODO: implement a wrapper/container component
-		// TODO: use the method in head/controller.php?
-		return get_the_title();
+	public function render_title(): string {
```
msawicki, line 12: I'm not sure I like `render_title()` as a name. I much prefer the previous `title()` or `get_title()` as it doesn't render anything unless it's echoed.

ryanurban: Fair enough, noting that this is something we should do a consistency pass on once we are wrapped with the getting everything working pivot work as I think this is what we or I have done thus far : )

PR #553, opened 2020-08-12
Feeture/share untwig

msawicki:
## What does this do/fix?

Adds Share component


wp-content/themes/core/components/share/share.php
```diff
@@ -0,0 +1,21 @@
...
+ * @var array $args Arguments passed to the template
+ */
+// phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UndefinedVariable
+$c = \Tribe\Project\Templates\Components\share\Controller::factory( $args );
+?>
+<aside class="social-share">
+
+	<h6 class="social-share__title"><?php esc_html_e( 'Share This', 'tribe' );?></h6>
+
+	<ul class="social-share-networks" data-js="social-share-networks">
```
msawicki, line 14: Wasn't sure if you want this stuff added as component options, same with header and the aside wrapper.

dpellenwood: It's fine for now. We'll need to make a a11y pass later. We can address anything that needs args then.

ryanurban: Yep, for starters we need to change `aside` to `div`, but can wait : )

wp-content/themes/core/components/content/single/index/index.php
```diff
@@ -30,7 +30,7 @@
 
 		</ul>
 
-		<?php //TODO: Share Component ?>
+		<?php get_template_part('components/share/share') ?>
```
ryanurban, line 33: This is just spiteful, but `get_template_part( 'components/share/share' );` 😂

PR #556, opened 2020-08-13
Comments - Untwig. 

msawicki:
## What does this do/fix?

Moves comment components over, adds comment pagination.
Add this to a single. Do we want to leave this here or should we remove it/comment it out since most projects don't have comments?


wp-content/themes/core/components/comments/comment/Controller.php
```diff
@@ -0,0 +1,130 @@
...
+			'reply_link' => '',
+			'time'       => [],
+		];
+	}
+
+	/**
+	 * @return array
+	 */
+	protected function required(): array {
+		$timestamp = get_comment_time( 'U' );
```
jbrinley, line 80: This controller is inconsistent about whether it's working with the global comment or a comment that has been passed in. I think we'll be better off if it standardizes on the global comment. A lot of the logic in the comment section controller could then be moved here.

wp-content/themes/core/components/comments/trackback/trackback.php
```diff
@@ -0,0 +1,23 @@
...
+$c = \Tribe\Project\Templates\Components\comments\trackback\Controller::factory( $args );
+?>
+
+<li <?php echo $c->classes(); ?><?php echo $c->attributes(); ?>>
+	<p>
+		<strong><?php echo esc_html( $c->label ); ?></strong>
+
+		<?php echo get_comment_author_link( $c->comment_id ); ?>
+
+		{% if edit_link %}
```
dpellenwood, line 17: Twig! :grin:

wp-content/themes/core/components/comments/comment/Controller.php
msawicki: Looked into this a bit more. While I agree, this is what I'm seeing. 
- The edit link uses a method to cut down on repeating code, if this moved to the trackback or comment controllers we'd repeat code that works here.
- There is a bit of a limitation with WP core function calls, some require a comment id and don't work inside the global comment scope without it so it's necessary. 
- The gravatar, this can be moved directly to the template but i'd guess we may want to use an image component here in the future. 

I'm not seeing much more that can be moved around. Wondering if you could provide some additional direction if you've got it?

jbrinley: I've pushed an update to break apart the single comment logic from the comment section logic. Does this seem like a reasonable direction?

wp-content/themes/core/components/pagination/comments/Comments_Pagination_Controller.php
```diff
@@ -0,0 +1,100 @@
...
+
+use Tribe\Libs\Utils\Markup_Utils;
+use Tribe\Project\Templates\Components\Abstract_Controller;
+
+class Comments_Pagination_Controller extends Abstract_Controller {
+	public const CLASSES = 'classes';
+	public const ATTRS   = 'attrs';
+	public const PAGED   = 'paged';
+
+	private array $classes;
```
msawicki, line 14: are we doing typed properties now?

wp-content/themes/core/components/comments/comment/Controller.php
msawicki: This is helpful. I'll probably wait to merge this until your changes from other branches get merged in so we can do some additional cleanup on the args stuff.

wp-content/themes/core/components/pagination/comments/Comments_Pagination_Controller.php
jbrinley: I don't know. I kind of like them. That's one of the things we can discuss here. Thoughts?

PR #557, opened 2020-08-13
Feeture/sqone 233/tabs block

dpellenwood:
## What does this do/fix?
Handles the Tab Component, Tabs Block, and related scripts & styles.


wp-content/plugins/core/src/Blocks/Types/Tabs/Tabs_Model.php
```diff
@@ -0,0 +1,67 @@
...
+	 */
+	private function get_title(): string {
+		return $this->get( Tabs_Block::TITLE, '' );
+	}
+
+	/**
+	 * @return string
+	 */
+	private function get_description(): string {
+		return apply_filters( 'the_content', $this->get( Tabs_Block::DESCRIPTION, '' ) ); // TODO: Unsure why this get() doesn't apply the_content automatically.
```
dpellenwood, line 47: @msawicki notice my comment here. Curious to hear your thoughts on it.

wp-content/plugins/core/src/Blocks/Types/Tabs/Tabs_Model.php
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	/**
+	 * @return array
+	 */
+	private function get_tabs(): array {
+		$tab_objects = [];
+		$tabs_data   = get_field( Tabs_Block::TABS ); // TODO: This doesn't appear to work: `$this->get( Tabs_Block::TABS, [] );` It returns the number of rows for the repeater as an integer.
```
dpellenwood, line 55: @msawicki Similarly here.

wp-content/themes/core/assets/js/src/theme/config/options.js
```diff
@@ -1,4 +1,5 @@
 
 // breakpoint settings
 
-export const MOBILE_BREAKPOINT = 768;
+export const MEDIUM_BREAKPOINT = 768;
```
dpellenwood, line 4: @ryanurban notice I'm adjusting the default breakpoints in our JS to better align with the way we do breakpoints in CSS. This has always been a new-dev pain point.

wp-content/plugins/core/src/Blocks/Types/Tabs/Tabs_Model.php
msawicki: Wondering if it does now with the recent changes to return the `get_field()` value from ACF?

wp-content/plugins/core/src/Blocks/Types/Tabs/Tabs_Model.php
msawicki: It should now if you pull from the epic branch. I think I merged this change in yesterday.

wp-content/themes/core/components/tabs/Tabs_Controller.php
```diff
@@ -0,0 +1,244 @@
...
+			self::TAB_BUTTON_CLASSES => [ 'c-tabs__tab' ],
+			self::TAB_PANEL_CLASSES  => [ 'c-tabs__tabpanel' ],
+		];
+	}
+
+	/**
+	 * Loop through the tabs provided setup the tab button and tab panel components arguments for each.
+	 */
+	private function init_tabs() {
+		foreach ( $this->tabs as $index => $tab ) {
```
msawicki, line 92: Interested in your thoughts on how I approached accordion. I setup the header_id and content_ids in the constructor for the Accordion Model. https://github.com/moderntribe/square-one/blob/epic/fidgety-feet/wp-content/plugins/core/src/Blocks/Types/Accordion/Accordion_Model.php#L36-L37 Wondering if we don't want to adopt something similar to what you have done here if there is anything specific that needs added other than a unique id.

wp-content/themes/core/components/tabs/Tabs_Controller.php
```diff
@@ -0,0 +1,244 @@
...
+	 *
+	 * @param Tab_Model $tab
+	 * @param string    $tab_id
+	 * @param int       $index
+	 *
+	 * @return array
+	 */
+	private function get_tab_button_args( Tab_Model $tab, string $tab_id, int $index ): array {
+		$args = [
+			'content' => $tab->label ?: sprintf( __( 'Tab %d', 'tribe' ), $index + 1 ),
```
msawicki, line 110: Escape $tab->label? I don't think it's escaped anywhere else down the line.

wp-content/themes/core/components/tabs/Tabs_Controller.php
```diff
@@ -0,0 +1,244 @@
...
+	 *
+	 * @param Tab_Model $tab
+	 * @param string    $tab_id
+	 * @param int       $index
+	 *
+	 * @return array
+	 */
+	private function get_tab_panel_args( Tab_Model $tab, string $tab_id, int $index ): array {
+		$args = [
+			'content' => $tab->content ?: '&nbsp;', // If the tab content is empty, the container component won't render.
```
msawicki, line 139: same comment as above, escaping on `$tab->content`?

wp-content/themes/core/components/tabs/Tabs_Controller.php
msawicki: I think maybe we need to do a pass on escaping in general.

wp-content/themes/core/components/tabs/Tabs_Controller.php
```diff
@@ -0,0 +1,244 @@
...
+	 *
+	 * @return array
+	 */
+	public function get_dropdown_toggle_args(): array {
+		if ( $this->layout === self::LAYOUT_HORIZONTAL ) {
+			return [];
+		}
+
+		return [
+			'content' => $this->tabs[0]->label ?? __( 'Tab 1', 'tribe' ),
```
msawicki, line 168: escape `$this-tabs[0]->label`

wp-content/plugins/core/src/Blocks/Types/Tabs/Tabs_Model.php
dpellenwood: Yep. Looks like it's working great now. Thanks!

wp-content/plugins/core/src/Blocks/Types/Tabs/Tabs_Model.php
dpellenwood: This as well. :rockout_more:

wp-content/themes/core/components/tabs/Tabs_Controller.php
dpellenwood: Yep. I like this arch a bit better because it focuses all the setup for the repeater things together in once place rather than passing generated uuids along with the "real" data into the component.

wp-content/themes/core/components/tabs/Tabs_Controller.php
dpellenwood: What difficult about this is I can see times where we'll pass arbitrary html into a button (a span, an icon SVG or <i>, etc).  So, what should we allow here? Maybe just use `wp_kses_post()`?

wp-content/themes/core/components/tabs/Tabs_Controller.php
dpellenwood: This for sure will contain html content, so similar question to above: what should be escaped versus what should be allowed?

PR #559, opened 2020-08-13
Feeture/hero block

msawicki:
## What does this do/fix?

Adds Hero Block


wp-content/plugins/core/src/Blocks/Types/Hero/Model.php
```diff
@@ -32,23 +25,36 @@ private function get_layout() {
...
 	/**
 	 * @return array
 	 */
-	private function get_content(): array {
+	private function get_cta_args(): array {
+		$cta = wp_parse_args( $this->get( Hero::CTA, [] ), [
+			'title'  => '',
+			'url'    => '',
+			'target' => '',
+		] );
```
ryanurban, line 40: Dumb q, what is the purpose of this?

msawicki: if the CTA is for some reason the default (an empty array) it will add some default array keys. Saves me from doing a bunch of `isset` calls or `??`

ryanurban: Ahh, cool, good idea

wp-content/themes/core/components/blocks/hero/Controller.php
```diff
@@ -61,61 +60,93 @@ class Controller extends Abstract_Controller {
...
+			] ),
+			'title'   => defer_template_part( 'components/text/text', null, [
+				'content' => $content[ Hero::TITLE ] ?? '',
+			] ),
+			'content' => defer_template_part( 'components/container/container', null, [
+				'content' => $content[ Hero::DESCRIPTION ] ?? '',
+			] ),
+			'cta'  => defer_template_part( 'components/link/link', null, $content[ Hero::CTA ] ?? [] ),
+			'layout'  => $this->layout,
+		] );
```
ryanurban, line 136: @msawicki thinking...give we want these block patterns as components to be usable outside of the context of ACF, we probably don't want to be baking in ACF field names here right?

wp-content/plugins/core/src/Blocks/Types/Hero/Model.php
```diff
@@ -32,23 +25,36 @@ private function get_layout() {
...
-			self::TEXT    => $this->get( Hero::DESCRIPTION, '' ),
-			self::CTA     => wp_parse_args( $this->get( Hero::CTA, [] ), [
-				'title'  => '',
-				'url'    => '',
-				'target' => '',
-			] ),
+			Hero::LEAD_IN     => $this->get( Hero::LEAD_IN, '' ),
+			Hero::TITLE       => $this->get( Hero::TITLE, '' ),
+			Hero::DESCRIPTION => $this->get( Hero::DESCRIPTION, '' ),
+			Hero::CTA         => $this->get_cta_args(),
```
ryanurban, line 57: Which also means we either need to ensure when we spin up new blocks that these constants map to the component args (and likely hoist are commonly used field groups like for the block header) OR default to using the straight component strings as the keys and only leverage the constants for the `get`. Thoughts?

cc @dpellenwood @jbrinley

ryanurban: My vote would be the latter so there is no confusion (along with creating commonly used field groups)

msawicki: I'm with you on this. I cleaned it up a bit. I'm making the component take specific arguments for each part of the content-block instead of an array of data.

PR #566, opened 2020-08-16
Unwrap Link Component

jbrinley:
## What does this do/fix?

* Removes the wrapper arguments from the button component. Calls to the button component that need a wrapper (none currently exist) should use the container component around it.
* Refactors the button controller class to meet the coding conventions that we've been discussing.
* Refactors calls to the link component to use constant references instead of magic strings.

## Tests

Does this have tests?

- [ ] Yes
- [X] No, this doesn't need tests because...this does not add new functionality

### DIFF
--- a/wp-content/themes/core/components/pagination/loop/Controller.php
+++ b/wp-content/themes/core/components/pagination/loop/Controller.php
@@ -5,6 +5,7 @@ namespace Tribe\Project\Templates\Components\pagination\loop;
 
 use Tribe\Libs\Utils\Markup_Utils;
 use Tribe\Project\Templates\Components\Abstract_Controller;
+use Tribe\Project\Templates\Components\link\Link_Controller;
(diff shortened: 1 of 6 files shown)


wp-content/themes/core/components/pagination/loop/Controller.php
```diff
@@ -127,9 +128,9 @@ public function build_links(): array {
 			}
 
 			return [
-				'classes' => $classes,
-				'url'     => $number['url'],
-				'content' => $number['label'],
+				Link_Controller::CLASSES => $classes,
+				Link_Controller::URL     => $number['url'],
+				Link_Controller::CONTENT => $number['label'],
```
jbrinley, line 133: A nice side effect of using the constants to reference our component args: it makes it clear that we're building an array of component args, even when we're not explicitly calling the component in the same place.

PR #567, opened 2020-08-18
Interstitial Block

msawicki:
## What does this do/fix?

Adds Interstitial Block component


wp-content/themes/core/components/blocks/interstitial/Interstitial_Block_Controller.php
```diff
@@ -0,0 +1,202 @@
...
+	public const CONTENT           = 'content';
+	public const CTA               = 'cta';
+	public const CONTAINER_CLASSES = 'container_classes';
+	public const MEDIA_CLASSES     = 'media_classes';
+	public const CONTENT_CLASSES   = 'content_classes';
+	public const CLASSES           = 'classes';
+	public const ATTRS             = 'attrs';
+
+	public string $layout;
+	public ?int $media;
```
ryanurban, line 30: typo?

msawicki: nullable

msawicki: In the instance that an image isn't provided, ACF will return an empty string. If it's provided, an int. So I'm setting default to null and allowing this property to be nullable or an int. the `?` is the way to set it as nullable.

ryanurban: Learn something new every day 😃

wp-content/themes/core/components/blocks/interstitial/Interstitial_Block_Controller.php
```diff
@@ -0,0 +1,202 @@
...
+	/**
+	 * @return array
+	 */
+	protected function defaults(): array {
+		return [
+			self::LAYOUT            => Interstitial_Block::LAYOUT_LEFT,
+			self::MEDIA             => null,
+			self::CONTENT           => '',
+			self::CTA               => [],
+			self::CONTAINER_CLASSES => [ 'b-interstitial__container', 'l-container' ],
```
dpellenwood, line 62: Hmm... some of these should be required classes. (I think this affected Hero & Interstitial too.)

wp-content/themes/core/components/blocks/interstitial/Interstitial_Block_Controller.php
```diff
@@ -0,0 +1,202 @@
...
+			self::MEDIA_CLASSES     => [ 'b-interstitial__media' ],
+			self::CONTENT_CLASSES   => [ 'b-interstitial__content' ],
+			self::CLASSES           => [ 'c-block', 'b-interstitial', 'c-block--full-bleed' ],
+			self::ATTRS             => [],
+		];
+	}
+
+	public function init() {
```
dpellenwood, line 70: Should we hoist `init()` to the abstract class? It seems like something we'll do for most Blocks.

jbrinley: I'm not particularly fond of having this `init` method anywhere. The name is not descriptive, and the functionality I've seen in it could just as easily move to the constructor. Or we standardize on a better named method like `set_vars()` and move all the var setting from the constructor to said method.

dpellenwood: I've been setting any arg-based classes/values within a respective get properties method, like so: https://github.com/moderntribe/square-one/compare/feeture/SQONE-233/tabs-block#diff-6d0b8db8917ed04a9b20faf6e3cdf7b0R117

jbrinley: ^ that is my preferred way of handling it.

PR #568, opened 2020-08-18
Quote Block

msawicki:
## What does this do/fix?

- Adds the quote block
- makes image fields return an id to be consistent with the typed properties
- Makes get_data() an abstract method. 
- renames Hero Model
- removes old quote block


wp-content/themes/core/components/blocks/quote/Quote_Block_Controller.php
```diff
@@ -0,0 +1,166 @@
...
+		$this->media_classes     = (array) $args[ self::MEDIA_CLASSES ];
+		$this->content_classes   = (array) $args[ self::CONTENT_CLASSES ];
+		$this->classes           = (array) $args[ self::CLASSES ];
+		$this->attrs             = (array) $args[ self::ATTRS ];
+	}
+
+	protected function defaults(): array {
+		return [
+			self::LAYOUT            => Quote_Block::MEDIA_OVERLAY,
+			self::MEDIA             => [],
```
jbrinley, line 60: `media` is declared as an `int` but has an array as the default value.

wp-content/themes/core/components/blocks/quote/Quote_Block_Controller.php
```diff
@@ -0,0 +1,166 @@
...
+	public const MEDIA_CLASSES     = 'media_classes';
+	public const CONTENT_CLASSES   = 'content_classes';
+	public const CLASSES           = 'classes';
+	public const ATTRS             = 'attrs';
+
+	public string $layout;
+	public ?int $media;
+	public string $cite_name;
+	public string $cite_title;
+	public ?int $cite_image;
```
jbrinley, line 34: Why not use `0` instead of `null` as a default empty value?

wp-content/themes/core/components/blocks/quote/Quote_Block_Controller.php
```diff
@@ -0,0 +1,166 @@
...
+	public string $quote_text;
+	public array $container_classes;
+	public array $media_classes;
+	public array $content_classes;
+	public array $classes;
+	public array $attrs;
```
jbrinley, line 40: We should discuss whether properties should be public or private.

wp-content/themes/core/blocks/hero/hero.php
```diff
@@ -1,5 +1,5 @@
 <?php
 declare( strict_types=1 );
-
-$model = new \Tribe\Project\Blocks\Types\Hero\Model( $args['block'] );// phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UndefinedVariable
+// phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UndefinedVariable
```
dpellenwood, line 3: Can you add [a PHPDoc above this](http://p.tri.be/EtRbLP) so that PHPStorm doesn't highlight `$args` as undefined: http://p.tri.be/uMJSIq?

wp-content/themes/core/blocks/quote/quote.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+declare( strict_types=1 );
+
+use Tribe\Project\Blocks\Types\Quote\Quote_Model;
+
+// phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UndefinedVariable
```
dpellenwood, line 7: Same here re: PHPDoc

wp-content/themes/core/blocks/quote/quote.php
```diff
@@ -0,0 +1,14 @@
...
+$model = new Quote_Model( $args[ 'block' ] );
+
+get_template_part(
```
dpellenwood, line 10: Dumb, but I'm curious why you're wrapping to new lines with this when it's shorter than ~80 chars/line length limits.

wp-content/themes/core/components/blocks/quote/Quote_Block_Controller.php
```diff
@@ -0,0 +1,166 @@
...
+
+	public function __construct( array $args = [] ) {
+		$args                    = $this->parse_args( $args );
+		$this->layout            = $args[ self::LAYOUT ];
+		$this->media             = $args[ self::MEDIA ];
+		$this->cite_name         = $args[ self::CITE_NAME ];
+		$this->cite_title        = $args[ self::CITE_TITLE ];
+		$this->cite_image        = $args[ self::CITE_IMAGE ];
+		$this->quote_text        = $args[ self::QUOTE_TEXT ];
+		$this->container_classes = (array) $args[ self::CONTAINER_CLASSES ];
```
dpellenwood, line 50: Do we still need the `(array)` typecasting for these args?

wp-content/themes/core/components/blocks/quote/quote.php
```diff
@@ -0,0 +1,33 @@
...
+ */
+// phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UndefinedVariable
+$c = Quote_Block_Controller::factory( $args ); ?>
+
+<section <?php echo $c->classes(); ?> <?php echo $c->attrs(); ?>>
+	<div <?php echo $c->container_classes(); ?>>
+
+		<?php if ( $c->media ) { ?>
+			<div <?php echo $c->media_classes(); ?>>
+				<?php get_template_part(
```
dpellenwood, line 16: Same q here re: wrapped arguments for `get_template_part()`

wp-content/themes/core/components/quote/Quote_Controller.php
```diff
@@ -7,61 +7,45 @@
...
 
-	public function __construct( array $args = [] ) {
-		$args = wp_parse_args( $args, $this->defaults() );
-
-		foreach ( $this->required() as $key => $value ) {
-			$args[$key] = array_merge( $args[$key], $value );
-		}
+	private array $classes;
+	private array $attrs;
+	public string $quote_text;
```
dpellenwood, line 20: Prob should update these to private and add getters for them, yes?

wp-content/themes/core/components/blocks/quote/Quote_Block_Controller.php
jbrinley: Yes. It ensures that we don't get a type error if the wrong kind of args were passed in.

dpellenwood: Should we be applying this for all args then?

jbrinley: Yes, I think it's a good habit to be in.

msawicki: I've wondered that. Someone could still pass in an different value by error and this would help catch that. But do we want to enable that or throw and error down the road to enforce them setting the correct value? I'm thinking maybe the latter?

wp-content/themes/core/components/quote/Quote_Controller.php
msawicki: Yep, I was planning to revisit

wp-content/themes/core/components/blocks/quote/quote.php
msawicki: In this instance it's like 87 characters. In the others, it's out of habit.

wp-content/themes/core/components/blocks/quote/Quote_Block_Controller.php
jbrinley: The robustness principle pushes me to handle it safely in the controller. While I'd like to think that we'd catch the errors in dev/qa, I don't 100% trust our process to catch it every time. I'd rather know that it's save than break the page in production.

PR #570, opened 2020-08-19
Feeture/Stats block

jamiepaul:
## What does this do/fix?

* Ports theme files for stats block and statistic component from previous work #524 
* Untwig stats block and statistic components
* Adds Stats block model

Given I haven't been a part of pivot work until yesterday, how'd I do? Pointers on where I'm not following conventions?

When I originally built the stats block, I had set up a statistic component. I'm not convinced it needs to be its own component. @dpellenwood and I discussed it here: https://github.com/moderntribe/square-one/pull/524#discussion_r463096167 @ryanurban thoughts? If we keep it as-is, stat items are currently rendering as Array Array, so I'm not sure I'm using `defer_component_part` correctly, where the text component is passed to the statistic component.


wp-content/themes/core/components/blocks/stats/Stats_Block_Controller.php
```diff
@@ -0,0 +1,178 @@
...
+	public string $title;
+	public string $description;
+	public array $container_classes;
+	public array $content_classes;
+	public array $classes;
+	public array $attrs;
+	public array $stats;
+
+	public function __construct( array $args = [] ) {
+		$args = wp_parse_args( $args, $this->defaults() );
```
msawicki, line 39: We have a method in Abstract_Controller that does all this now. We can do `$args = $this->parse_parts($args);`

wp-content/themes/core/components/blocks/stats/Stats_Block_Controller.php
```diff
@@ -0,0 +1,178 @@
...
+			self::CLASSES           => [ 'c-block', 'b-stats' ],
+			self::ATTRS             => [],
+			self::STATS             => [],
+		];
+	}
+
+	/**
+	 * @return array
+	 */
+	protected function required(): array {
```
msawicki, line 78: We can remove this method since there isn't anything going on.

wp-content/themes/core/components/blocks/stats/Stats_Block_Controller.php
```diff
@@ -0,0 +1,178 @@
...
+		return [];
+	}
+
+	/**
+	 * Initial setup stuff
+	 */
+	public function init() {
+		$this->classes[] = 'c-block--' . $this->layout;
```
msawicki, line 86: We are ditching the init methods in favor of just doing this logic in the `classes()` method

wp-content/themes/core/components/statistic/Controller.php
```diff
@@ -0,0 +1,119 @@
...
+	 * @uses components/text
+	 */
+	private $value;
+	/**
+	 * @var Deferred_Component
+	 * @uses components/text
+	 */
+	private $label;
+
+	public const TAG     = 'tag';
```
msawicki, line 36: Couple of things.
- Let's move the constants above properties.
- We're naming Controllers to be more descriptive now. So `Statistics_Controller`
- We're doing typed properties, which is new in php 7.4. See https://github.com/moderntribe/square-one/blob/epic/fidgety-feet/wp-content/themes/core/components/blocks/logos/Logos_Block_Controller.php#L25-L33

wp-content/themes/core/components/statistic/Controller.php
```diff
@@ -0,0 +1,119 @@
...
+	public const CLASSES = 'classes';
+	public const ATTRS   = 'attrs';
+	public const VALUE   = 'value';
+	public const LABEL   = 'label';
+
+	public function __construct( array $args ) {
+		$args = wp_parse_args( $args, $this->defaults() );
```
msawicki, line 43: Same comment as above about `parse_args()`

wp-content/themes/core/components/statistic/Controller.php
```diff
@@ -0,0 +1,119 @@
...
+use Tribe\Project\Templates\Components\Deferred_Component;
+
+/**
+ * Class Statistic
+ */
+class Controller extends Abstract_Controller {
+	/**
+	 * @var string
+	 */
+	public $tag;
```
msawicki, line 16: As of this morning, we're making all our properties `private` by default and using `get_*` methods to get the values in the template. We are going to do a sweep to update these so don't worry about hitting them all right now. Let's just get this one done!

wp-content/plugins/core/src/Blocks/Types/Stats/Stats_Model.php
```diff
@@ -0,0 +1,21 @@
...
+class Stats_Model extends Base_Model {
+
+	public function get_data(): array {
+		return [
+			Stats_Block_Controller::LAYOUT           => $this->get( Stats::LAYOUT, Stats::LAYOUT_STACKED ),
+			Stats_Block_Controller::CONTENT_ALIGN    => $this->get( Stats::CONTENT_ALIGN, Stats::CONTENT_ALIGN_LEFT ),
+			Stats_Block_Controller::DISPLAY_DIVIDERS => $this->get( Stats::DISPLAY_DIVIDERS, true ),
+			Stats_Block_Controller::TITLE            => $this->get( Stats::TITLE, '' ),
+			Stats_Block_Controller::DESCRIPTION      => $this->get( Stats::DESCRIPTION, '' ),
+			Stats_Block_Controller::STATS            => $this->get( Stats::STATS, [] ),
```
dpellenwood, line 18: I ran into an issue getting repeater values this way in the Tabs Block. It returned the repeater count, rather than the array of values.

Regardless, it might be nice to setup a "Statistic Template Model" here so that the data provided to the Block and respective Components has the correct format/values. Here's a set of examples from the Tabs Block:
* You can see how I set that up in the Tabs component [here](https://github.com/moderntribe/square-one/pull/557/files#diff-438d1ba3de385ebf852fd24c46543ea3R53-R65).
* The Tab Template Model is [here](https://github.com/moderntribe/square-one/blob/epic/fidgety-feet/wp-content/plugins/core/src/Templates/Models/Tab.php).
* Then, you can see I'm denoting that the incoming argument type should be an array of those Tab Models [here](https://github.com/moderntribe/square-one/pull/557/files#diff-421120537c1c477ee0b791f1dc53d242R25-R28).
* Finally, you can see how I reference each Model Object's properties (rather than needing to reference array keys) [here](https://github.com/moderntribe/square-one/pull/557/files#diff-421120537c1c477ee0b791f1dc53d242R139) (and elsewhere in this file).

wp-content/plugins/core/src/Blocks/Types/Stats/Stats.php
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	public function add_fields() {
+		$this->add_field( new Field( self::NAME . '_' . self::TITLE, [
+				'label' => __( 'Title', 'tribe' ),
+				'name'  => self::TITLE,
+				'type'  => 'text',
+			] )
+		)->add_field(
+			new Field( self::NAME . '_' . self::DESCRIPTION, [
```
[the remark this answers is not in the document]
ryanurban: Agreed, that makes sense to me!

wp-content/plugins/core/src/Blocks/Types/Stats/Stats_Model.php
msawicki: @dpellenwood this should be corrected with a recent update to the Base_Model on how we are getting fields. It shouldn't return the number of repeaters any longer.

wp-content/plugins/core/src/Blocks/Types/Stats/Stats.php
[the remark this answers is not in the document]
jamiepaul: done and done 👍

wp-content/themes/core/components/blocks/stats/Stats_Block_Controller.php
```diff
@@ -0,0 +1,162 @@
...
+			self::DISPLAY_DIVIDERS  => Stats_Block::DISPLAY_DIVIDERS,
+			self::TITLE             => '',
+			self::DESCRIPTION       => '',
+			self::CONTAINER_CLASSES => [ 'b-stats__container', 'l-container' ],
+			self::CONTENT_CLASSES   => [ 'b-stats__content' ],
+			self::CLASSES           => [ 'c-block', 'b-stats' ],
+			self::ATTRS             => [],
+			self::STATS             => [],
+		];
+	}
```
ryanurban, line 69: Dave may have already called this out so forgive me, but I'd set defaults for the 3 class args to be empty arrays and then utilize the `required()` method and implement what you have there as required base classes

wp-content/themes/core/components/blocks/stats/Stats_Block_Controller.php
```diff
@@ -0,0 +1,162 @@
...
+	public function content_classes(): string {
+		return Markup_Utils::class_attribute( $this->content_classes );
+	}
+
+	/**
+	 * @return string
+	 */
+	public function classes(): string {
+		$this->classes[] = 'c-block--' . $this->layout;
+		$this->classes[] = 'c-block--' . $this->display_dividers;
```
ryanurban, line 151: Unsure of the above class, but I'd go `b-stats--` vs. `c-block--` as the base prefix for this particular class as it is very contextually based on the stats block vs. a globally needed block thing

wp-content/themes/core/components/statistic/Statistic_Controller.php
```diff
@@ -0,0 +1,99 @@
...
+	public const CLASSES = 'classes';
+	public const ATTRS   = 'attrs';
+	public const VALUE   = 'value';
+	public const LABEL   = 'label';
+
+	public string $tag;
+	private array $classes;
+	private array $attrs;
+	private $value;
+	private $label;
```
jamiepaul, line 23: @msawicki I got statistics to render (yay) but I had to remove the array type here, I think they should be Deferred_Component? Tried a few ways of doing that and decided PHP is not my first language.

wp-content/themes/core/components/blocks/stats/Stats_Block_Controller.php
```diff
@@ -0,0 +1,179 @@
...
+	 */
+	public function get_stats(): array {
+		$statistic_args = [];
+
+		foreach ( $this->stats as $item ) {
+			// Skip over statistic rows with no value.
+			// TODO: fix fatal error caused by calling $item[ Stats_Block::ROW_VALUE ].
+//			if ( empty( $item[ Stats_Block::ROW_VALUE ] ) ) {
+//				continue;
+//			}
```
jamiepaul, line 113: @msawicki I'm trying to skip the entire row if there's no value set for the repeater's ROW_VALUE field but it's causing a fatal when I'm checking the condition like this (http://p.tri.be/V4zQ2C) is there a better way to do this?

wp-content/themes/core/components/blocks/stats/Stats_Block_Controller.php
jamiepaul: added this back w/ the class args from defaults

wp-content/themes/core/components/blocks/stats/Stats_Block_Controller.php
```diff
@@ -0,0 +1,178 @@
...
+
+		return $statistic_args;
+	}
+
+	/**
+	 * @return array
+	 */
+	protected function get_value_args( $item ): array {
+		return [
+			'content' => esc_html( $item[ Stats_Block::ROW_VALUE ] ),
```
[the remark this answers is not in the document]
jamiepaul: @dpellenwood you raise a good point, I could for sure see `<sup>` or `<sub>` maybe being needed here. To handle the `<sup><sub>` cases, maybe the `get_value_args()` method is modified to include the inner html needed next to the value and it then returns `defer_template_part()` instead of an array of args?

For wrapping `<span>` or another element around the text component, I don't see as many use cases.. at that point wouldn't we just modify the statistic component itself to have the wrapper? I'm not well versed in how y'all are handling components needing containers, so maybe there's an easier way to make this flexible that I'm not seeing.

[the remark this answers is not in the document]
dpellenwood: For now, you can leave this as-is, imho. I think we'll just have to figure this out as we go. Thanks, Jamie!

wp-content/themes/core/components/blocks/stats/Stats_Block_Controller.php
dpellenwood: Imho, I think we should skip this check and just render the empty stat. An editor would have to intentionally _add_ a stat/row without adding any values to it. Maybe they want it blank for a reason. And, if they don't then they can easily just remove the row in the editor. (Just my 2-cents.)

PR #608, opened 2020-08-29
Feature/gha deploy

nickpelton:
## What does this do/fix?

Updates deployments to use GitHub Actions instead of our traditional Ansible or Jenkins systems. Why?
* It makes deployments accessible the any developer who has access to the repository
* It's relatively easy for anyone to use as it's YAML and easy to understand
* With the `workflow_dispatch` event we can manually deploy any branch as needed. Even QA folks can deploy a branch.
* Eliminates bot deploys, which has alway been a mystery to most of the team.
* Included with github, new and shiny, etc.

## QA

- [Link to Demo](https://squareone2.wpengine.com/)

Screenshots/video:
- [Link to successful deploy](https://github.com/moderntribe/square-one/actions/runs/229484458)

## Tests

Does this have tests?

- [x] No, this doesn't need tests because it's a Deploy system written in YAML.


nickpelton:
Once this is approved for merge, I'd like to setup the server branches again for proper deploys.

nickpelton:
Tests now still pass.

.github/workflows/deploy-stage.yml
```diff
@@ -0,0 +1,133 @@
...
+# Install composer: Add env for plugin keys, composer install, and cleanup
+      - name: Install composer dependencies
+        working-directory: ${{ env.BUILD_FOLDER }}
+        run: |
+          echo "${{ secrets.COMPOSER_ENV }}" > .env
+          composer install --optimize-autoloader --ignore-platform-reqs --no-dev
+          rm .env
+          rm -rf vendor/moderntribe/tribe-libs/.git
+          rm -rf wp-content/plugins/panel-builder/.git
+          rm -rf wp-content/plugins/gutenpanels/.git
```
msawicki, line 43: I don't think wee need this one.

.github/workflows/deploy-stage.yml
```diff
@@ -0,0 +1,133 @@
+name: Staging Deploy
+
```
msawicki, line 2: One thing we should include here in either a README or a doc block are all the secrets you'll need to set. It can be challenging to scan these files and figure out what needs set.

marcossantiago:
I honestly prefer the scripts to avoid repetition and even help local builds. But approved.

marcossantiago (approved):

.github/workflows/deploy-stage.yml
nickpelton: removed.

.github/workflows/deploy-stage.yml
nickpelton: Added docs.

nickpelton:
> I honestly prefer the scripts to avoid repetition and even help local builds. But approved.

I agree. But as I built this out, I found it more complicated to manage the scripts when it's really just a few lines of code. So I chose simplicity over DRY. Open to refining in the future.

PR #610, opened 2020-08-31
Feeture/post list block

msawicki:
## What does this do/fix?

Adds Post List Block

I had to make a couple changes to the card component that are included in this.


wp-content/plugins/core/src/Blocks/Types/Post_List/Post_List.php
```diff
@@ -0,0 +1,213 @@
+<?php
```
ryanurban, line 1: Nice, I'll likely just turn this into the "Content" block when we get there. I'm wondering if we should turn this "field" into something more modular perhaps where it could be easier to deploy as needed for any given block?

wp-content/plugins/core/src/Blocks/Types/Post_List/Post_List.php
```diff
@@ -0,0 +1,213 @@
...
+				'toolbar'      => 'basic',
+				'media_upload' => 0,
+			] )
+		)->add_field( new Field( self::NAME . '_' . self::CTA, [
+				'label' => __( 'Call to Action', 'tribe' ),
+				'name'  => self::CTA,
+				'type'  => 'link',
+			] )
+		)->add_field( new Field( self::NAME . '_' . self::QUERY_TYPE, [
+				'label'   => __( 'Type of Query', 'ur-core' ),
```
ryanurban, line 80: usage of `ur-core` to `tribe` through out

wp-content/plugins/core/src/Blocks/Types/Post_List/Post_List.php
msawicki: I'm not 100% sure I'm following. Do you mean breaking out the Post Query functionality so it can be re-used in other blocks?

wp-content/plugins/core/src/Blocks/Types/Post_List/Post_List.php
msawicki: 🍝

wp-content/plugins/core/src/Blocks/Types/Post_List/Post_List.php
ryanurban: Sorry didn't explain it well. Yes, basically turning this overall post list field itself, which currently consists of multiple fields, either into one field or one partial to more easily be deployed across blocks as you know it will be.

msawicki: Ah yes, 100%. I'd like to not rely on third-party fields if we don't have to and roll our own but ⏲️

ryanurban: Fair enough, I'll spin up a ticket for it for later

PR #617, opened 2020-09-03
Feature/acf block class names

dpellenwood:
## What does this do/fix?

Adds support for WP Core's [Additional Classes](http://p.tri.be/nJNqxw) field to be applied to our ACF blocks.

## Tests

Does this have tests?

- [ ] Yes

### DIFF
--- a/wp-content/plugins/core/src/Blocks/Types/Accordion/Accordion_Model.php
+++ b/wp-content/plugins/core/src/Blocks/Types/Accordion/Accordion_Model.php
@@ -14,6 +14,7 @@ class Accordion_Model extends Base_Model {
 	 */
 	public function get_data(): array {
 		return [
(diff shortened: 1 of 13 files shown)


ryanurban (approved):

jbrinley (commented):
Functionally, this looks fine. I would prefer we name things differently.

1. We use snake_case in our PHP code. While recognizing that the data coming in from the block uses "className", we are under no obligation to maintain that formatting in our code.
2. I always feel like "class name" is redundant. The name _is_ the class. `$this->classes` seems just as meaningful as `$this->class_names`.

msawicki (approved):
I agree on the `snake_case` changes. I'm indifferent on the `class_name` vs `classes` change suggested. Thanks for doing this.!

PR #618, opened 2020-09-03
Update block/component block editor admin overrides implementation

ryanurban:
@dpellenwood let me know what you think. I def took some liberties here. If you like this approach, if you could also confirm that the visual still holds up with a quick glance... 😂 ❤️


wp-content/themes/core/assets/css/src/admin/block-editor/alignments.pcss
```diff
@@ -0,0 +1,77 @@
...
+ * by the block editor and wp-admin.
+ *
+ * ----------------------------------------------------------------------------- */
+
+.block-editor-block-list__layout {
+
+	.wp-block-audio,
+	.wp-block-video,
+	.wp-block-embed,
+	figure.wp-block-image {
```
ryanurban, line 16: @dpellenwood considered doing the same `block-editor.pcss` approach for the `css/theme/media/` and the other more sink base level implementations for consistency. Thoughts?

dpellenwood (approved):
Looks great. Thanks for tackling it.

wp-content/themes/core/assets/css/src/admin/block-editor/alignments.pcss
dpellenwood: Yes, I think that would be a nice addition as well.

PR #619, opened 2020-09-04
Disable object cache when running tests

jbrinley:
## What does this do/fix?

Our base test suite runs fine with object cache enabled. But once you
try to use it on multisite, we start to see conflicts between the
acceptance suite and the webdriver suite, with differing results
depending on the order the tests run.

By setting the environment variable DISABLE_OBJECT_CACHE, we prevent
the object cache drop-in from fully loading when tests are running.

Note that this is likely the first step towards moving our object
cache file out to tribe-libs, where we could maintain separate
packages for different caching backends.

## Tests

Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests because...the whole point is to configure something to _not_ run during tests.


msawicki (approved):
Thanks!

PR #622, opened 2020-09-04
Feature/block previews

dpellenwood:
## What does this do/fix?
Adds the required data for Gutenberg to display a preview of each ACF block. The preview is available when browsing for block with the block inserter sidebar.

## QA
Links:
- [Developer Documentation](https://developer.wordpress.org/block-editor/developers/block-api/block-registration/#example-optional)
- [Demo](http://p.tri.be/oGueNW)

## Tests
Does this have tests?
- [ ] Yes
- [x] No, this doesn't need tests because it's just demo content.


ryanurban (approved):
Noting that design will do a pass on this as well (have been discussing with them as a thing). Good catch!

PR #624, opened 2020-09-07
Port tribe branding plugin bits into core plugin

ryanurban:
This retires our use of the Tribe Branding plugin. We'll now leverage WP's Site Icon setting and then I ported over the additional settings that I think are worth keeping into the core plugin.

@jbrinley @binarygary def make sure I did this right. Thanks!


ryanurban:
@jbrinley I need some help with broken tests here when you have a moment this week. Feel free to reach out and walk me through this for the future. Thanks!

wp-content/plugins/core/src/Integrations/Site_Brand/Site_Brand_Subscriber.php
```diff
@@ -0,0 +1,34 @@
+<?php
+declare( strict_types=1 );
+
+namespace Tribe\Project\Integrations\Site_Brand;
```
jbrinley, line 4: What plugin is this integrating with? Just as in the theme, the intention of the "Integrations" namespace is to contain code that integrates our project with 3rd-party plugins.

ryanurban: Good point...I wondered about this. Perhaps this belongs in the `Admin` directory or a new `Branding` directory, @jbrinley?

jbrinley: I like `Branding`, either top-level or under `Theme`.

wp-content/plugins/core/src/Theme/Branding/Customizer_Settings.php
```diff
@@ -0,0 +1,64 @@
...
+		) );
+
+		// IE Metro Icon Background Color
+		$wp_customize->add_setting( self::SITE_BRANDING_IE_BGD_COLOR );
+		$wp_customize->add_control( new WP_Customize_Color_Control(
+			$wp_customize,
+			self::SITE_BRANDING_IE_BGD_COLOR,
+			[
+				'label'       => __( 'IE Metro Icon Background Color', 'tribe' ),
+				'description' => __( 'Select the background color to be used behind the IE Metro Icon.', 'tribe' ),
```
dpellenwood, line 57: I think we should rename this to something more up-to-date/relevant and change the instructions to be clearer. 
1. IE is dead
2. This is primarily for site's "pinned" to your Taskbar in Windows, yes?

dpellenwood (approved):

wp-content/plugins/core/src/Theme/Branding/Customizer_Settings.php
ryanurban: Yep, good point. This is for Windows 8 / IE tile bgd color, which is no longer supported by us or Microsoft 😀 Removed it.

jbrinley (approved):

PR #630, opened 2020-09-10
Feature/ui ux tribe blocks

ryanurban:
Need to get this merged, fixes and cleans up a slew of block issues


wp-content/plugins/core/src/Blocks/Types/Post_List/Post_List.php
```diff
@@ -41,8 +32,8 @@ class Post_List extends Block_Config {
 	 */
 	public function add_block() {
 		$this->set_block( new Block( self::NAME, [
-			'title'       => __( 'Post List', 'tribe' ),
-			'description' => __( 'A block of curated posts', 'tribe' ),
+			'title'       => __( 'Post List Field', 'tribe' ),
+			'description' => __( 'A first pass on panel\'s post list field', 'tribe' ),
```
msawicki, line 36: I'd recommend a different description here. Something that doesn't tell the history that this used to be a panel. 
"A block to curate posts manually or by building a dynamic query" or something similar?

msawicki (approved):
Looks like a lot of changes to Post List that I assume are ok/based on design. One comment on the description of the Block.

wp-content/plugins/core/src/Blocks/Types/Post_List/Post_List.php
ryanurban: Post List for now is just an example of the post list field. This will get turned into a legit fully functioning block post design pass, hopefully next week or so.

PR #631, opened 2020-09-11
Remove Image model

jbrinley:
## What does this do/fix?

* Removes the `Image` and `Image_Derivative` models.
* Replaces the `attachment` argument to `Image_Controller` with `img_id`

## Tests

Does this have tests?

- [x] Yes, this adds minimal tests to ensure that srcset sizes are loaded from attachment meta.


ryanurban (approved):
🔥

wp-content/themes/core/components/blocks/hero/Hero_Block_Controller.php
```diff
@@ -232,6 +232,7 @@ public function get_image_args(): array {
 		return [
 			Image_Controller::IMG_ID       => $this->media,
 			Image_Controller::AS_BG        => true,
+			Image_Controller::AUTO_SHIM    => false,
```
ryanurban, line 235: I'm still planning to revisit our image component in the future. I think we can simplify our args at this point and make them a bit more intuitive in additional to refining and cleaning up our overall logic in the image component.

PR #634, opened 2020-09-15
SQONE-498 Add Tota11y plugin for dev & testing. 

dpellenwood:
## What does this do/fix?
Adds the WP Tota11y plugin for dev & QA. We used to inject the script as an NPM package and separate JS equeue as part of the SquareOne build. This was removed when the build system was refactored as part of FF.  QA has asked that we bring it back as it's useful for QA'ing. And, frankly, it's helpful for dev as well.

I'm proposing we just rely on a 3rd-party plugin for this rather than making it a part of the Sq1 build system. As it is setup in this PR, the plugin will be force-deactivated if WP_DEBUG is false (or Production), and it will be available to network admins to activate only (in multisite).

## Tests
Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests because it's just a plugin addition.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -102,7 +102,8 @@
     "wpackagist-plugin/safe-svg": "1.9.9",
     "wpackagist-plugin/the-events-calendar": "5.1.6",
     "wpackagist-plugin/user-switching": "1.5.5",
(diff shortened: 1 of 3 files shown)


jbrinley (approved):
🕶️

PR #637, opened 2020-09-24
Post List UX Changes

msawicki:
## What does this do/fix?

UX Improvements to Post List
- Changes Field Config
-- Uses button group for type
-- Makes manual posts selection a repeater
-- Adds fields for creating content
-- Uses slider for limit
- Updates Post_List_Object to follow more of a builder pattern, this allows us to change values when manually overriding with custom contents.

## Demo Videos: 
- [My Walkthrough](http://p.tri.be/Lk9Fyf)
- [Kyle's Walkthrough](https://www.loom.com/share/296c14407e9046118204dce36424d0b4)


ryanurban (approved):
Nice, next step is turning this into an easier to reuse across blocks field!

wp-content/plugins/core/src/Blocks/Types/Post_List/Post_List.php
```diff
@@ -48,110 +61,210 @@ public function add_fields() {
...
-							'field'    => 'field_ ' . self::NAME . '_' . self::QUERY_TYPE,
-							'operator' => '==',
-							'value'    => self::QUERY_TYPE_AUTO,
-						],
-						[
-							'field'    => 'field_' . self::NAME . '_' . self::POST_TYPES,
-							'operator' => '==',
-							'value'    => $post_type,
+							'field'    => 'field_' . self::NAME . '_' . self::QUERY_TAXONOMIES,
+							'operator' => '==contains',
```
jbrinley, line 239: Is "==contains" a real operator? Or is this supposed to be "contains"?

wp-content/plugins/core/src/Templates/Models/Post_List_Object.php
```diff
@@ -11,72 +11,135 @@ class Post_List_Object {
...
+
+	/**
+	 * @var array
+	 */
+	private $link;
+
+	/**
+	 * @var string
+	 */
+	private $post_type;
```
jbrinley, line 47: What was the reasoning behind removing all the type declarations?

dpellenwood (approved):
This looks great, Mike. Thank you for tackling it!

dpellenwood:
Let me know if we need to handle some admin styling here as well.

wp-content/plugins/core/src/Blocks/Types/Post_List/Post_List.php
msawicki: Oh it's real. Because ACF.

wp-content/plugins/core/src/Templates/Models/Post_List_Object.php
msawicki: Great question. Because of the uninitialized state, when I check if the object "is valid" the get_title() methods throw errors. I felt this was cleaner/easier? Wondering your thoughts? Maybe the getters do something like this
'''
public function get_title() {
    return isset($this-title) ? $this-title : '';
}
'''

defunctl: Why not initialize the variables when they're declared, e.g. `private string $title = '';` ?

msawicki: Yep, good idea. It's not a pattern we've followed for the components is the only reason I didn't do this but this is a separate thing.

defunctl: Don't forget to add the return types back to the methods :)

jbrinley (approved):

PR #638, opened 2020-09-24
Update configs

ryanurban:
Updates browserslist and a couple of items for our dev configs


dpellenwood (approved):
:real_heart: Thanks for doing this. Will be nice to have those starter files all cleaned up.

PR #639, opened 2020-09-24
Feature/card grid block

dpellenwood:
## What does this do/fix?

Builds out the base card grid block, pre "card" component finishes. I had to have a system to use cards before I could style them. Note that there are some changes here that I know Mike has made edits to in #637 and Ryan in #638. I'd like to propose we merge those before we merge this.

[SQONE-440]

[SQONE-440]: https://moderntribe.atlassian.net/browse/SQONE-440


wp-content/plugins/core/src/Blocks/Types/Card_Grid/Card_Grid_Model.php
```diff
@@ -0,0 +1,152 @@
...
+			'target' => '',
+		] );
+
+		return [
+			Link_Controller::CONTENT => $cta[ 'title' ],
+			Link_Controller::URL     => $cta[ 'url' ],
+			Link_Controller::TARGET  => $cta[ 'target' ],
+		];
+	}
+
```
ryanurban, line 42: From here on down, sure would be nice to have some post list fields models or helper methods so we don't have to copy/paste this every time we want to nab the data for the field, though it does give us the ultimate flexibility to tweak and customize too.

wp-content/themes/core/components/blocks/card_grid/css/card_grid.pcss
```diff
@@ -0,0 +1,38 @@
...
+
+/* -----------------------------------------------------------------------------
+ * Cards Loop
+ * ----------------------------------------------------------------------------- */
+
+.b-card-grid__loop {
+	display: grid;
+	grid-gap: var(--grid-gutter-small);
+	grid-template-columns: repeat(auto-fit, minmax(min(var(--grid-3-col), 100%), var(--grid-4-col)));
+	/*grid-template-columns: repeat(auto-fill, minmax(var(--grid-3-col), var(--grid-4-col)));*/
```
ryanurban, line 28: Is this needed? Perhaps a doc note if so?

ryanurban (approved):
🔥

wp-content/plugins/core/src/Blocks/Types/Card_Grid/Card_Grid_Model.php
dpellenwood: Yep! I know that's on the short list of enhancements.

wp-content/themes/core/components/blocks/card_grid/css/card_grid.pcss
dpellenwood: Good catch. It is not needed at all. Removed.

PR #640, opened 2020-09-24
Feature/component card

dpellenwood:
## What does this do/fix?
Handles the base card component implementation and related scripts and styles. Note that this is based from the card grid block PR #639 and shouldn't be merged until that does.

[SQONE-191]

[SQONE-191]: https://moderntribe.atlassian.net/browse/SQONE-191


wp-content/themes/core/assets/css/src/theme/layouts/grids.pcss
```diff
@@ -0,0 +1,121 @@
+/* -----------------------------------------------------------------------------
+ *
+ * Grids
```
dpellenwood, line 3: Took a stab at some general grid layouts here for use throughout. If this is the wrong approach, I'm happy to pull these in favor of applying it directly in the card grid block.

wp-content/themes/core/assets/css/src/theme/spacing/aspect-ratios.pcss
```diff
@@ -0,0 +1,33 @@
+/* -------------------------------------------------------------------------
+ *
+ * Aspect Ratio Helpers
+ *
+ * ------------------------------------------------------------------------- */
+
+.s-aspect-21-9 {
```
dpellenwood: I've added these AR helpers under the spacing namespace as it seemed like the most relevant application.

wp-content/themes/core/components/blocks/post_list/Post_List_Controller.php
```diff
@@ -85,28 +87,58 @@ public function get_posts_card_args() {
...
+						Link_Controller::URL     => $link['url'],
+						Link_Controller::CLASSES => [ 'a-cta', 'is-target-link' ],
+						Link_Controller::ATTRS   => [
+							// These attrs provide the most screen reader accessible link.
+							'id'               => $uuid . '-link',
+							'aria-labelledby'  => $uuid . '-title',
+							'aria-describedby' => $uuid . '-link',
+							// Sets this link as the card's click-within target link.
+							'data-js'          => 'target-link',
+						],
```
ryanurban, line 141: Nothing to change, just thinking out loud...There is enough needed for this particular type of card from the a11y front + link functionality piece that especially if we go the card component per card variation route, it might make more sense to set these as defaults vs needing to declare them every time.

One other thought regarding the target linking, to make it simpler vs. needing to also define the things at every level, I was considering turning that into more of a card level bool argument that then flips on/off the right classes/etc. for the whole card. Looks like you have the bool already and are using it that way to some degree. Thoughts about leveraging it to rely 100% on the arg?

ryanurban (approved):
Solid 💪

wp-content/themes/core/components/blocks/post_list/Post_List_Controller.php
dpellenwood: I thought about that as well. My concern is that a card could not use the CTA field and/or have another link (such as the post title) that should be used as the target link. In the spirit of keeping the low-level components as flexible as possible, I think it makes more sens to let the usage dictate any application of the values/settings and options.

PR #642, opened 2020-09-26
Update .editorconfig pcss rule

alexdelgado:
The previously defined glob for .pcss files did not resolve properly the editor would not apply editor config settings.

## What does this do/fix?

Updates the glob for `.pcss` files.

## QA

Links to relevant issues
- [Link to Issue](https://example.com)

- [Link to Demo](https://example.com)

Screenshots/video:

## Tests

Does this have tests?

- [x] Yes

- Open your editor and verify that `.pcss` files use tabs instead of spaces when auto-formatted.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -26,8 +26,8 @@ spaces_within_imports = true
 indent_style = space
 indent_size = 2
 
-# Indentation override for all CSS under pcss directory
-[wp-content/themes/core/pcss/**/**/*.pcss]
+# Indentation override for all CSS under theme directory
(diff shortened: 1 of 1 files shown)


jbrinley (approved):
Good catch; things moved around in our recent rearchitecture and we didn't update the `.editorconfig`. We should probably do this for the JS, too.

ryanurban:
Good catch ❤️

ryanurban (approved):
You can go ahead and merge this!

PR #644, opened 2020-09-30
Update 1password link in .env.sample

defunctl:
## What does this do/fix?

The 1password link moved and we'll now use a short URL we can adjust in the future in the event that happens again.

## Tests

Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests because it's a comment change.


jbrinley (approved):
Good idea to have a redirect we can maintain. How do we maintain it?

defunctl:
> Good idea to have a redirect we can maintain. How do we maintain it?

Nick P has direct database access to change it in the future :)

PR #645, opened 2020-10-01
Update .editorconfig js rule

alexdelgado:
## What does this do/fix?

The previously defined glob for .js files did not resolve properly so the editor would not apply editor config settings.

## Tests

Does this have tests?

- [x] Yes: Open your editor and verify that .js files use tabs instead of spaces when auto-formatted.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -16,8 +16,8 @@ indent_size = 2
 [*.md]
 trim_trailing_whitespace = false
 
-# Indentation override for all JS under assets directory
-[wp-content/themes/core/assets/js/src/**/*.js]
+# Indentation override for all JS under theme directory
+[wp-content/themes/core/**/*.js]
 indent_style = tab
 spaces_within_imports = true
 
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -16,8 +16,8 @@ indent_size = 2
 [*.md]
 trim_trailing_whitespace = false
 
-# Indentation override for all JS under lib directory
-[wp-content/themes/core/js/src/**/*.js]
+# Indentation override for all JS under assets directory
+[wp-content/themes/core/assets/js/src/**/*.js]
```
jbrinley, line 20: I vote for changing this to `wp-content/themes/core/**/*.js`, since we may have JS in the `components` and `integrations` directories, too.

ryanurban: 💯 on what Jonathan said and we do indeed

ryanurban (approved):
Thanks, Alex! See Jonathan's note and then this is gtg 🏅

jbrinley (approved):

PR #646, opened 2020-10-01
Updates fields to use field section accordions

msawicki:
## What does this do/fix?

Per the conversation this week, this change used the tribe-libs new Field_Section to create content sections inside the plugin and out of tribe libs. This is specifically using accordions for the sections. This is dependent on the changes in https://github.com/moderntribe/tribe-libs/pull/69

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,10 @@ jobs:
       COMPOSE_FILE: docker-compose.yml:docker-compose.gh.yml
 
     steps:
+      - name: Setup PHP, with composer and extensions
+        uses: shivammathur/setup-php@v2 #https://github.com/shivammathur/setup-php
+        with:
+          tools: composer:v1
 
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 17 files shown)


ryanurban (commented):
Thanks for hitting this too, Mike! In this shift I'm not sure we need even a formal `Field_Section` abstraction either. It's one less thing we'd need to translate from design via gui acf declarations just using the accordion field. Thoughts?

msawicki:
Note the tests are failing because of the missing dependency in tribe libs.

msawicki:
@ryanurban I feel pretty strongly about having the separation here for sections. It's a nice visual and structural breakup of the section definition and personally think it improves the overall readability and ability to see what's going on. It is, however, optional. I dev could just use ->add_field() for a section with the ACF config array.

ryanurban:
Fair enough and true

dpellenwood (approved):
Looks great. Thank you for tackling this.

jbrinley (approved):
I like this API much better.

dpellenwood:
@msawicki is there anything holding this up from merging?

PR #648, opened 2020-10-08
Add correct image width & height attributes for SVG images.

dpellenwood:
## What does this do/fix?
The Safe SVG plugin sets the width & height attributes for SVG images to false. Our image lazy-loading library complaints about this with console errors. This PR re-filters the attributes for SVG images to set a valid width & height value based on the requested size and the image's original width & height values.

Of note: this code won't play well if the SVG is supposed to be cropped (like other image formats) for a particular size request. Instead, all size requests are scaled (with aspect ratio integrity). This is fine any SVG usage I can think of, but MMV.

## Tests
Does this have tests?
- [ ] Yes

### DIFF
--- a/wp-content/plugins/core/src/Theme/Media/SVG_Filters.php
+++ b/wp-content/plugins/core/src/Theme/Media/SVG_Filters.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace Tribe\Project\Theme\Media;
+
+class SVG_Filters {
+	/**
+	 * Filters the image src result.
+	 *
+	 * The Safe SVG plugin sets the width & height attributes to false.
+	 * This causes our FE srcset values to be invalid for lazysizes.
+	 *
+	 * WordPress saves the correct SVG image dimensions in the image's full-size attributes.
+	 * We're going to use the full-size values and set the correct values based on the
+	 * requested size's aspect ratio.
+	 *
+	 * Note: This code will ALWAYS scale (not crop) the SVG to fit the requested size.
+	 * If your use-case requires cropping SVG images, then this filter should be disabled.
+	 *
+	 * @param array|false  $image         Either array with src, width & height, icon src, or false.
+	 * @param int          $attachment_id Image attachment ID.
+	 * @param string|array $size          Size of image. Image size or array of width and height values (in that
+	 *                                    order). Default 'thumbnail'.
+	 * @param bool         $icon          Whether the image should be treated as an icon. Default false.
+	 *
+	 * @return array
+	 * @filter wp_get_attachment_image_src
+	 */
+	public function set_accurate_sizes( $image, $attachment_id, $size, $icon ) {
+		if ( get_post_mime_type( $attachment_id ) !== 'image/svg+xml' ) {
+			return $image;
+		}
+
+		$meta = wp_get_attachment_metadata( $attachment_id );
+
+		if ( is_array( $size ) ) {
+			// If a specific width & height are requested, just use them
+			$requested_size['width']  = $size[0];
+			$requested_size['height'] = $size[1];
+		} else {
+			// Otherwise check for the requested size key in the available sizes.
+			$requested_size = isset( $meta[ 'sizes' ][ $size ] ) ? $meta[ 'sizes' ][ $size ] : false;
+
+			if ( ! $requested_size ) {
+				return $image;
+			}
+		}
+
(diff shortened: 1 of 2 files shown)


ryanurban (approved):

dpellenwood:
@jbrinley would love your input on this PR as well, if you'd be so kind.

jbrinley (commented):
1. I would like to move this to https://github.com/moderntribe/tribe-libs/tree/master/src/Media
2. I would like to add some tests.
3. There is a possibility for division by zero that we'll want to avoid.

wp-content/plugins/core/src/Theme/Media/SVG_Filters.php
```diff
@@ -0,0 +1,61 @@
...
+	 * @return array
+	 * @filter wp_get_attachment_image_src
+	 */
+	public function set_accurate_sizes( $image, $attachment_id, $size, $icon ) {
+		if ( get_post_mime_type( $attachment_id ) !== 'image/svg+xml' ) {
+			return $image;
+		}
+
+		// Don't process images existing size with values
+		if ( $image[1] !== false && $image[2] !== false ) {
```
jbrinley, line 34: I was trying to write some tests for this, and `$image[1]` and `$image[2]` always had values. As you say in the comment above, Safe SVG is setting these to `false`; the values are properly set by WP beforehand.

Here's where it happens in the Safe SVG code: https://plugins.trac.wordpress.org/browser/safe-svg/trunk/safe-svg.php#L238

I'm wondering, @dpellenwood, why don't we disable this filter from Safe SVG? That seems simpler than allowing it to run and then undoing it later.

1. Are the values we get from WP (before Safe SVG strips them out) accurate?
2. If these values are not accurate, how? Could we correct them when WP generates image metadata instead of recalculating every time we render it?

jbrinley: > If these values are not accurate, how?

It looks like the Safe SVG plugin is generating the size metadata when the image is uploaded/regenerated. It sets the sizes at whatever the image size definition says the size should be. So your `square-xsmall` will be 150 x 150, your `sixteen-nine` will be 1280 x 720, and your `core-full` will be 1600 x 0.

The image will then be scaled to fit the box, with whitespace added around the image to fill it out. And that `core-full` image will end up 1px tall.

Here's what I propose:

1. We let Safe SVG generate the size meta as it does now.
2. After it does so, we hook in and fix any values that need to be fixed.
3. We disable the filter on `wp_get_attachment_image_src` in Safe SVG so that it stops stripping out the dimensions. WP will then work with the sizes we specified in step 2.

Now let's add more clarity around step 2.
2.1. If a dimension is 0, we calculate the appropriate value based on the original aspect ratio.
2.2. If a dimension is 9999, we treat it as a zero (see 2.1). This is somewhat common to find in WP plugins.
2.3. If an image size is set to _not_ crop, we calculate the appropriate aspect ratio to fit within the image dimensions.
2.4. If an image size _is_ set to crop, we set the dimensions to whatever is defined for that image size, knowing that this will result in padding around our image to fill whichever direction needs filling. (We could set a filter on this behavior so a project could easily change it when appropriate.)

What are your thoughts, @dpellenwood? @ryanurban?

jbrinley: As I review the Safe SVG code some more, I think we need to rewrite the plugin. Did you know, for example, that it's loading the image file to memory multiple times every time the image is rendered, just so it can re-calculate the dimensions of the SVG?

dpellenwood: Thanks for such a thorough review and deep dive, Jonathan. I have no "attachment" ;) to this plugin or any other. We just need a way to safely upload and use SVG images while retaining the expected metadata & functionality of other image types in WordPress.

I think your proposed solution (specifically step 2, re: resizing) sounds quite reasonable. Weather we tackle the solution via filtering the current plugin, rewrite it completely, or use a different solution I'm in agreement without on how sizes should be handled.

How can I best assist this effort?

dpellenwood:
Closing in favor of Jonathan's work in Tribe Libs: https://github.com/moderntribe/tribe-libs/pull/71

ryanurban:
@dpellenwood are we good to delete this branch at this point?

PR #650, opened 2020-10-14
adds ability to remove core block styles

ChrisMKindred:
## What does this do/fix?

Adds a consistent location to remove the core block styles that are registered client-side since they [cannot be removed](https://developer.wordpress.org/block-editor/developers/filters/block-filters/#unregister_block_style) using the server-side `unregister_block_style` function.

## Tests

Does this have tests?

- [ ] Yes
- [X] No, I need help figuring out how to write the tests.


ryanurban (approved):
Nice, thanks for adding this Chris, I'll need this for the upcoming UI/UX design pass on core/common blocks!

PR #656, opened 2020-11-04
Force deployments to use composer v1 and don't yarn validate

msawicki:
## What does this do/fix?

This should get deploys running again, forcing use of composer v1 and skipping `yarn validate` for now.

### DIFF
--- a/.github/workflows/deploy-dev.yml
+++ b/.github/workflows/deploy-dev.yml
@@ -18,6 +18,10 @@ jobs:
 
       ##########
       ### BUILD: General SquareOne build steps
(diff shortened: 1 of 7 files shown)


jbrinley (approved):
Can we just delete `yarn validate` instead of commenting it?

nickpelton (approved):
👍

If yarn validate is failing should we have the UI team fix that?

dpellenwood:
I can fix the linter issues here on this branch. Please hold...

dpellenwood:
@msawicki `yarn validate` should no longer fail on deploys.

PR #658, opened 2020-11-09
Use 74-2.1 to ensure composerv1

defunctl:
## What does this do/fix?

Use the minor version of the docker image to continue to get patch updates. `74-2.1` would pull the version that forces composer v1.

## QA

None.

## Tests

Does this have tests?

- [ ] Yes
- [ x] No, this doesn't need tests because it's a configuration update

### DIFF
--- a/dev/docker/docker-compose.yml
+++ b/dev/docker/docker-compose.yml
@@ -32,7 +32,7 @@ x-chrome: &chrome
 
 
 x-php: &php
-  image: moderntribe/squareone-php:74-2.1.1
+  image: moderntribe/squareone-php:74-2.1
   working_dir: /application
   user: "${SQ1_UID:-1000}:${SQ1_GID:-1000}"
   volumes:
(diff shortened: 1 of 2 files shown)


dev/docker/docker-compose.yml
```diff
@@ -32,7 +32,7 @@ x-chrome: &chrome
 
 
 x-php: &php
-  image: moderntribe/squareone-php:74-2.1.1
+  image: moderntribe/squareone-php:74-2.1
```
jbrinley, line 35: Is this the right order? It looks like we're going backwards.

defunctl: Yeah, it will always grab `74-2.1.x`: https://hub.docker.com/layers/moderntribe/squareone-php/74-2.1/images/sha256-10762c1dd18858d8a7f59c3bf1efa2871b922c3154822b2b14e3d92e0b65634b?context=explore

jbrinley (approved):

PR #659, opened 2020-11-13
Adds core/block to allow list

ChrisMKindred:
## What does this do/fix?

Adds the core/block to the allow list to provide reusable block functionallity.

## Tests

Does this have tests?

- [ ] Yes
- [X] No, I need help figuring out how to write the tests.


jbrinley (approved):
I would like to sort this list of allowed blocks alphabetically, but that thought has nothing to do with this PR.

jbrinley (approved):
Thanks!

PR #661, opened 2020-12-03
Content Loop Block

ScottGarrison:
## What does this do/fix?

Adds the Content Loop block into the common block setup for SQ1 Gutenberg 

Based on DaveE's advice I branched `epic/grey-goose` off of master so we can merge and deploy to dev without messing with `master`

## BE Notes

- I am not a BE dev. This could use some strict review.
- In content-loop.php there are some things I did to handle the different layout options of feature/row/columns. Let me know if you have a better implementation of how I can achieve this.

## FE Notes

- I added a new card-horizontal format since it needed to be switched back and forth in the same layout. It was the cleanest way I could think of but might need to be changed depending on BE feedback.
- I updated the grid layout because it put negative margins on the top and bottom, which seems uneeded.

image
image
image


wp-content/themes/core/components/blocks/content_loop/index.js
```diff
@@ -0,0 +1,16 @@
+/* -----------------------------------------------------------------------------
+ *
+ * Component: Content Loop
+ *
+ * This file is just a clearing-house, see the css directory
+ * and edit the source files found there.
+ *
+ * ----------------------------------------------------------------------------- */
+
+import content_loop from './js/content_loop';
```
dpellenwood, line 10: Hmm... Will there be scripts specific to this module? If not, these files can be deleted.

wp-content/themes/core/components/blocks/content_loop/css/content_loop.pcss
```diff
@@ -0,0 +1,147 @@
+/* -----------------------------------------------------------------------------
+ *
+ * Content Loop Header
```
dpellenwood, line 3: General comment: It would be nice to add header and `/* CASE: ... */` comments for all the style blocks ([like so](https://github.com/moderntribe/square-one/blob/main/wp-content/themes/core/components/blocks/card_grid/css/card_grid.pcss#L9-L14)).

wp-content/themes/core/components/card/Card_Controller.php
```diff
@@ -25,10 +25,11 @@ class Card_Controller extends Abstract_Controller {
...
 
-	public const STYLE          = 'style';
-	public const STYLE_PLAIN    = 'plain';
-	public const STYLE_ELEVATED = 'elevated';
-	public const STYLE_OUTLINED = 'outlined';
+	public const STYLE            = 'style';
+	public const STYLE_PLAIN      = 'plain';
+	public const STYLE_ELEVATED   = 'elevated';
+	public const STYLE_OUTLINED   = 'outlined';
+	public const STYLE_HORIZONTAL = 'horizontal';
```
dpellenwood, line 32: The design call this `inline`, but I'm fine with horizontal.

wp-content/themes/core/components/card/card-horizontal.php
```diff
@@ -0,0 +1,47 @@
...
+d.bind_events = function() {
+	d.$el.body.on("click", ".use-target", function() {
+		a.location = c(this).find(".is-target").attr("href")
+	}).on(d.state.click, ".save-target", function(a) {
+		a.stopPropagation()
+	}),
+};
+card.style.cursor = 'pointer';
+also scope stanford 125
+*/
```
dpellenwood, line 20: Let's delete all this. Card a11y is already baked in.

wp-content/themes/core/components/card/card-horizontal.php
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * @var array $args Arguments passed to the template
+ */
+// phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UndefinedVariable
+$c = Card_Controller::factory( $args );
+?>
+
+<<?php echo $c->get_tag(); ?> <?php echo $c->get_horizontal_classes(); ?> <?php echo$c->get_attrs(); ?>>
```
dpellenwood, line 29: Curious if it's really necessary to have a separate template file for this card style? It looks like it just omits the description? If that's the case, I'd almost rather see us support the description in this card style too. Then in a project, if description is unused, it just doesn't render.  (fwiw, the designs [show that the larger horizontal card](the designs show that the larger horizontal card can show a description.) can show a description.)

wp-content/themes/core/components/card/css/card.pcss
```diff
@@ -119,3 +119,50 @@
...
+ * Component: Card Horizontal
+ *
+ * ----------------------------------------------------------------------------- */
+
+.c-card--style-horizontal {
+	display: flex;
+	flex-flow: row;
+	align-items: center;
+
+	.c-card__title {
```
dpellenwood, line 134: Please add these styles as a conditional selector within the existing element selectors (like CTA on lines 113-121 above). And please add CASE comments as needed.

dpellenwood (changes requested):
hey Scott, fantastic work here. I know I've left a lot of comments. The vast majority are based on making sure the code stays consistent between blocks/developers. Definitely nothing  wrong with how you're approaching it, we just want to make sure things stay the same from block to block.  Let me know if it would be helpful to jump on a call and chat through some of the bigger changes. I'm here to help.

wp-content/plugins/core/src/Blocks/Blocks_Definer.php
```diff
@@ -30,6 +30,7 @@ class Blocks_Definer implements Definer_Interface {
 	public function define(): array {
 		return [
 			self::TYPES => DI\add( [
+				DI\get( Types\Content_Loop\Content_Loop::class ),
```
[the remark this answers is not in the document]
ScottGarrison: I can. This was actually added in automatically with JB's block creation script. I can look into adding feedback about it there also.

ScottGarrison:
@dpellenwood Thanks for the review! Made a bunch of updates to this if you could take another look.

wp-content/plugins/core/src/Blocks/Types/Content_Loop/Content_Loop.php
```diff
@@ -0,0 +1,325 @@
...
+					'label' => __( 'Description', 'tribe' ),
+					'name'  => self::DESCRIPTION,
+					'type'  => 'wysiwyg',
+				] )
+			)->add_field( new Field( self::NAME . '_' . self::CTA, [
+					'label' => __( 'Call to Action', 'tribe' ),
+					'name'  => self::CTA,
+					'type'  => 'link',
+				] )
+			)->add_field( new Field( self::NAME . '_' . self::QUERY_TYPE, [
```
dpellenwood, line 102: I wonder if we should try to land the ACF query field work Mike did as part of this block. Thoughts, @ryanurban ?

wp-content/themes/core/assets/css/src/theme/typography/_mixins.pcss
```diff
@@ -210,3 +210,17 @@
 		}
 	}
 }
+
+/* -------------------------------------------------------------------------
+ * Tag
+ * ------------------------------------------------------------------------- */
+
+@define-mixin t-tag {
+	display: inline-block;
```
dpellenwood, line 219: super minor, but there's a white space issue here. A quick `gulp lint` in the theme directory should solve this and any other little things.

wp-content/themes/core/components/blocks/content_loop/css/content_loop.pcss
```diff
@@ -0,0 +1,134 @@
...
+
+        @media (--viewport-medium) {
+            margin-right: var(--spacer-40);
+            margin-top: 0;
+        }
+    }
+
+    @media (--viewport-medium) {
+        display: grid;
+        grid-template-columns: 66.6666% 33.3333%;
```
dpellenwood, line 21: Nice! Pumped to see some grid usage here!

wp-content/themes/core/components/card/css/card.pcss
```diff
@@ -60,6 +82,23 @@
...
 	}
+
+	/* CASE: Style: Inline */
+	.c-card--style-inline & {
+		display: none;
+		flex-shrink: 0;
+
+		@media(--viewport-medium) {
+			display: block;
+			width: 152px;
```
dpellenwood, line 93: Does this align to a grid width? ANy chance we can use a grid width variable? Or, if not, can you document why it's specific? Same with line 99 below.

wp-content/themes/core/components/blocks/content_loop/css/content_loop.pcss
```diff
@@ -0,0 +1,134 @@
...
+        .c-card:nth-child(1),
+        .c-card:nth-child(2) {
+            border-top: 1px solid var(--color-grey-light);
+        }
+    }
+
+    /* CASE: Feature */
+    &.b-content-loop--layout_feature {
+        /* hide content block cta on this layout */
+        .c-content-block__cta {
```
dpellenwood, line 80: If the `c-content-block__cta` selector above was un-nested (top-level) than this conditional application (and the one in the header) could both be added as conditions inside the single instance of the cta selector.

wp-content/themes/core/components/blocks/content_loop/css/content_loop.pcss
```diff
@@ -0,0 +1,134 @@
...
+                }
+
+                @media (--viewport-full) {
+                    width: calc(50% - var(--grid-gutter));
+                    margin: 0 var(--grid-gutter-half);
+                }
+            }
+        }
+
+        .b-content-loop__secondary {
```
dpellenwood, line 106: Does this and the `.b-content-loop__secondary__cta` selector need to be nested in `.b-content-loop`?  Again here, I'd prefer to see these as top-level selectors with conditionals inside them for any differences.

dpellenwood (approved):
Some asks related to nested selectors for the block, both otherwise super clean. Thanks for tackling this, Scott!

wp-content/themes/core/components/blocks/content_loop/css/content_loop.pcss
ScottGarrison: I agree with this one, but the one in the header feels better to keep next to all the other elements defined in the `grid-tepmlate-areas`

wp-content/plugins/core/src/Blocks/Types/Content_Loop/Content_Loop.php
ScottGarrison: I'm going to let someone figure that out in another branch/ticket

PR #665, opened 2020-12-15
Gallery Grid Block

ScottGarrison:
## What does this do/fix?

Adds in the Gallery Grid Block to SQ1

- Gives the ability to pull images used into a slideshow
- Settings allow option to choose column layout 1 through 4 up.


wp-content/plugins/core/src/Blocks/Blocks_Definer.php
```diff
@@ -30,6 +30,7 @@ class Blocks_Definer implements Definer_Interface {
 	public function define(): array {
 		return [
 			self::TYPES => DI\add( [
+				DI\get( Types\Gallery_Grid\Gallery_Grid::class ),
```
dpellenwood, line 33: Would you kindly import (`use ...`) the namespace and then class the class name only here?

wp-content/plugins/core/src/Blocks/Types/Gallery_Grid/Gallery_Grid_Model.php
```diff
@@ -0,0 +1,26 @@
...
+use Tribe\Project\Templates\Components\blocks\gallery_grid\Gallery_Grid_Controller;
+use Tribe\Project\Templates\Components\link\Link_Controller;
+
+/**
+ * Class Gallery_Grid_Model
+ *
+ * Responsible for mapping values from the block to arguments
+ * for the component
+ */
+class Gallery_Grid_Model extends \Tribe\Project\Blocks\Types\Base_Model {
```
dpellenwood, line 15: Similar here, `(use ... )` and just the class name here.

wp-content/themes/core/assets/js/src/theme/core/components.js
```diff
@@ -11,6 +11,7 @@ import share from 'components/share';
 import slider from 'components/slider';
 import tabs from 'components/tabs';
 import video from 'components/video';
+import gallery from 'components/blocks/gallery_grid';
```
dpellenwood, line 14: Let's keep the name name consistent here (`blockGalleryGrid` or similar), since there's a good chance we'll have other "gallery" modules to import.

wp-content/themes/core/components/blocks/gallery_grid/Gallery_Grid_Controller.php
```diff
@@ -0,0 +1,346 @@
...
+
+	/**
+	 * Get the Slider
+	 *
+	 * @return array
+	 */
+	public function get_slider_args(): array {
+		$main_attrs = [];
+
+		$swiper_options = '{"preloadImages":"false","lazy":"true","watchSlidesVisibility":"true","spaceBetween":60,"ally":"true","keyboard":"true","grabCursor":"true","navigation":{"nextEl":".swiper-button-next","prevEl":".swiper-button-prev"}}';
```
dpellenwood, line 207: Can you adjust this to use a PHP array to define the options, then use `json_encode()` to encode them into the JSON string before sending it along? It makes reading or editing the options / config much easier later.  [Here's an example from Conroe](https://github.com/ConroeISD/external-website/blob/sprint/2020.11/wp-content/plugins/core/src/Templates/Panels/MediaGallery.php#L232-L244) (uses a different slider lib, but the concept is the same.)

wp-content/themes/core/components/blocks/gallery_grid/Gallery_Grid_Controller.php
```diff
@@ -0,0 +1,346 @@
...
+		];
+
+		return $slider;
+	}
+
+	/**
+	 * @return array
+	 */
+	public function get_gallery_img_ids(): array {
+		$gallery_ids = [];
```
dpellenwood, line 227: I think there's probably a way to simplify this method a bit, but I'll leave it to a BE to suggest any changes. ;)

wp-content/themes/core/components/blocks/gallery_grid/Gallery_Grid_Controller.php
```diff
@@ -0,0 +1,346 @@
...
+		] );
+	}
+
+	/**
+	 * @param $img_id
+	 *
+	 * @return string
+	 */
+	protected function get_image_template( $index, $img_id ): string {
+		$img          = wp_get_attachment_image( $img_id, Image_Sizes::CORE_FULL );
```
dpellenwood, line 269: Hmm.... I'd prefer to see you use an actual image component here, along with a few image sizes (even just WP's `medium` & `large` in addition to the `CORE_FULL` size. That will better support mobile devices along with wrapping the slide content in a `figure` element.

Additionally, the Image component supports an `$html` property which can be used to add a `figcaption`, which is your HTML markup below.

wp-content/themes/core/components/blocks/gallery_grid/Gallery_Grid_Controller.php
```diff
@@ -0,0 +1,346 @@
...
+		$slide_markup = $img;
+		$slide_markup .= '
+			<div class="b-gallery-grid__meta-wrap">
```
dpellenwood, line 272: A few things here:
* Hard-coding the html markup here feels wrong. Would it be possible to use a container and text components here instead?
* For the count string, please use: `sprintf( __( '%d of %d', 'tribe' ), $index + 1, count( $this->get_gallery_img_ids() );` It's cleaner and easier to translate later.
* I think the `of` should be lower cased, even if it is styled differently in the designs.

wp-content/themes/core/components/blocks/gallery_grid/Gallery_Grid_Controller.php
```diff
@@ -0,0 +1,346 @@
...
+		$img_bg       = true;
+		$img_bg_class = 'c-image--bg';
+		$img_aspect   = 's-aspect-ratio-1-1';
+
+		if ( empty( $ids ) ) {
+			return $gallery_imgs;
+		}
+
+		if ( $this->grid_layout === self::ONE ) {
+			$img_size     = Image_Sizes::CORE_FULL;
```
dpellenwood, line 315: I worry that this, too, will be too large for mobile devices. I think we should use both a `src_size` size and `srcset_sizes` so we don't unintentionally load over-sizes images.

wp-content/themes/core/components/blocks/gallery_grid/css/dialog.pcss
```diff
@@ -0,0 +1,268 @@
...
+ *
+ * Component: Dialog
+ *
+ * ----------------------------------------------------------------------------- */
+
+.c-dialog {
+
+    .swiper-wrapper {
+        align-items: center;
+        padding: 20px 0;
```
dpellenwood, line 11: Please use the existing spacer values for any padding/margins (here and throughout).

wp-content/themes/core/components/blocks/gallery_grid/css/dialog.pcss
```diff
@@ -0,0 +1,268 @@
+/* -----------------------------------------------------------------------------
+ *
+ * Component: Dialog
+ *
+ * ----------------------------------------------------------------------------- */
+
+.c-dialog {
```
dpellenwood, line 7: Hmm.... this component's css is quite a collection of other component's styles. I think you need to more clearly define what is specific to a generic `dialog` and move the rest of the styles to other component's styles or in the case of the swiper styles, move those to that integration's styles.

wp-content/themes/core/components/blocks/gallery_grid/css/dialog.pcss
```diff
@@ -0,0 +1,268 @@
...
+        position: static;
+    }
+
+    .swiper-container-autoheight,
+    .swiper-container-autoheight .swiper-slide {
+        text-align: center;
+    }
+
+    .site-panel--gallery__col {
+        padding-bottom: 30px;
```
dpellenwood, line 32: Does this class exist in the blocks/components system? If it does, it should be renamed.

wp-content/themes/core/components/blocks/gallery_grid/css/dialog.pcss
```diff
@@ -0,0 +1,268 @@
...
+        font-size: 16px;
+    }
+}
+
+.c-dialog[aria-hidden="true"] {
+    display: none;
+}
+
+.c-dialog__overlay {
+    background-color: #151515;
```
dpellenwood, line 130: Please be sure to use existing color vars (here and throughout).

wp-content/themes/core/components/blocks/gallery_grid/css/dialog.pcss
```diff
@@ -0,0 +1,268 @@
...
+    bottom: 0;
+    color: var(--color-white);
+    height: 100%;
+    left: 0;
+    overflow: scroll;
+    position: fixed;
+    right: 0;
+    top: 0;
+    width: calc(100% + 20px);
+    z-index: 99998;
```
dpellenwood, line 140: This is unnecessarily high. Can we use something more reasonable?

wp-content/themes/core/components/blocks/gallery_grid/css/dialog.pcss
```diff
@@ -0,0 +1,268 @@
...
+    background-color: #151515;
+    bottom: 0;
+    color: var(--color-white);
+    height: 100%;
+    left: 0;
+    overflow: scroll;
+    position: fixed;
+    right: 0;
+    top: 0;
+    width: calc(100% + 20px);
```
dpellenwood, line 139: Curious why the overlay is 20px wider than 100%?

wp-content/themes/core/components/blocks/gallery_grid/css/dialog.pcss
```diff
@@ -0,0 +1,268 @@
...
+        top: var(--spacer-wpadmin-mobile);
+
+        @media (--viewport-wpadmin) {
+            top: var(--spacer-wpadmin-desktop);
+        }
+    }
+}
+
+.c-dialog__overlay-inner {
+    height: calc(100% - 68px);
```
dpellenwood, line 152: Please create a var for this since the value is reused a few places. it's easier to adjust a single var than have to remember to update values in multiple locations that should match.

wp-content/themes/core/components/blocks/gallery_grid/css/dialog.pcss
```diff
@@ -0,0 +1,268 @@
...
+            content: var(--icon-chevron-left);
+        }
+    }
+
+    .swiper-pagination-fraction {
+        font-size: 16px;
+    }
+}
+
+.c-dialog[aria-hidden="true"] {
```
dpellenwood, line 125: Can you add some CSS doc blocks to these selectors (here and throughout)? It makes grepping the code easier later.

wp-content/themes/core/components/blocks/gallery_grid/css/dialog.pcss
```diff
@@ -0,0 +1,268 @@
...
+    width: 100%;
+
+    @media (--viewport-full) {
+        width: 80%;
+        padding: 0 20px;
+    }
+}
+
+.c-dialog__close-button {
+    background-color: transparent;
```
dpellenwood, line 206: Imho, I think this should be abstracted to a unique "close" action style that we can use throughout the system rather than specific to the dialog here.

wp-content/themes/core/components/blocks/gallery_grid/css/dialog.pcss
```diff
@@ -0,0 +1,268 @@
...
+.b-gallery-grid__meta-count {
+    width: 25%;
+    text-align: right;
+    padding-right: 30px;
+}
+
+.b-gallery-grid__meta-caption {
+    flex: 1;
+    text-align: left;
+    max-width: 580px;
```
dpellenwood, line 267: This can use a `grid-cols` var.

wp-content/themes/core/components/blocks/gallery_grid/css/gallery_grid.pcss
```diff
@@ -0,0 +1,88 @@
+.b-gallery-grid__content {
```
dpellenwood, line 1: Can you add CSS doc blocks to this file?

wp-content/themes/core/components/blocks/gallery_grid/css/gallery_grid.pcss
```diff
@@ -0,0 +1,88 @@
...
+	}
+}
+
+/* -----------------------------------------------------------------------------
+ *
+ * Component: Gallery Grid
+ *
+ * ----------------------------------------------------------------------------- */
+
+.gallery-layout--one {
```
dpellenwood, line 29: Please go ahead and use CSS grid (and grid gutters) for these layouts. No need to use flexbox for grid layouts. :tada:

wp-content/themes/core/components/blocks/gallery_grid/gallery_grid.php
```diff
@@ -0,0 +1,64 @@
...
+ */
+// phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UndefinedVariable
+$c = \Tribe\Project\Templates\Components\blocks\gallery_grid\Gallery_Grid_Controller::factory( $args );
+?>
+
+<section <?php echo $c->get_classes(); ?> <?php echo $c->get_attributes(); ?>>
+	<div <?php echo $c->get_container_classes(); ?>>
+
+		<div class="b-gallery-grid__header">
+			<?php if ( ! empty( $c->get_header_args() ) ) { ?>
```
dpellenwood, line 15: Curious why you're wrapping the content block in another container? You can apply the `b-gallery-grid__header` class to the content block (also, would be good to make it a html `<header>` if it's not), and you can use the content blocks' CTA field as the slideshow button rather than appending it after the fact. (less markup for the same effect.)

wp-content/themes/core/components/blocks/gallery_grid/gallery_grid.php
```diff
@@ -0,0 +1,64 @@
...
+	
+	</div>
+</section>
+
+<?php if ( $c->is_slideshow() ) : ?>
+	<script data-js="dialog-content-<?php echo $c->get_block_id(); ?>" type="text/template">
+		<div class="c-dialog">
+			<div class="c-dialog__overlay">
+				<div class="c-dialog__header">
+					<div class="c-dialog__title"><?php echo $c->get_slideshow_title(); ?></div>
```
dpellenwood, line 48: Should the dialog title be a heading tag? Additionally, I think this should all be moved to a dialog component's template, and controller and rendered via a `get_template_part()` rather than hard coded in this component's template.

wp-content/themes/core/components/blocks/gallery_grid/index.pcss
```diff
@@ -0,0 +1,11 @@
+/* -----------------------------------------------------------------------------
+ *
+ * Component: Gallery Grid
+ *
+ * This file is just a clearing-house, see the css directory
+ * and edit the source files found there.
+ *
+ * ----------------------------------------------------------------------------- */
+
+@import "css/dialog.pcss";
```
dpellenwood, line 10: This should be imported as it's own component.

wp-content/themes/core/components/blocks/gallery_grid/js/gallery_grid.js
```diff
@@ -0,0 +1,96 @@
...
+ * @module
+ * @description JavaScript that drives Dialog
+ */
+
+import A11yDialog from 'mt-a11y-dialog';
+import * as tools from 'utils/tools';
+import { trigger } from 'utils/events';
+
+const el = {
+	containers: tools.getNodes( '[data-js="dialog-trigger"]', true, document, true ),
```
dpellenwood, line 11: This trigger is pretty generic. Either it should apply to all dialogs (and be init'd in the dialog component), or the trigger should be specific to the gallery grid block.

dpellenwood (changes requested):
A ton of great work here, Scott. Thanks for tackling this block.

There are some aspects of the code that need to be pulled a part some though. Specifically, I think you'll want to introduce a true, independent `dialog` component and separate the Gallery Grid block's code from the Dialog & Slider code a bit. As it it is right now, the components are too interwoven to be reusable (or editable) outside of this block's context.  Happy to jump on a call and talk it out.

wp-content/themes/core/components/blocks/gallery_grid/Gallery_Grid_Controller.php
mmcachran: return ! empty( $this->gallery ) ? array_filter( wp_list_pluck( $this->gallery, 'id' ) ) : [];

ScottGarrison:
@dpellenwood Large overhaul done on this based on your feedback. Please give another run through when you are available.

wp-content/themes/core/assets/css/src/theme/actions/_mixins.pcss
```diff
@@ -262,3 +262,40 @@
 		}
 	}
 }
+
+/* -----------------------------------------------------------------------------
+ * Button: Close
+ * ----------------------------------------------------------------------------- */
+
+@define-mixin a-close {
```
dpellenwood, line 270: Some white space issues here. A `gulp lint` in the theme directory before merging should ix this and any other small issues throughout.

wp-content/themes/core/components/blocks/gallery_grid/Gallery_Grid_Controller.php
```diff
@@ -0,0 +1,421 @@
...
+			$img_bg       = false;
+			$img_bg_class = '';
+			$img_aspect   = '';
+			$img_srcset   = [ 
+				'medium',
+				'medium_large',
```
dpellenwood, line 371: Are `medium` and `medium_large` actual WP Core sizes? I wasn't aware that they existed.

wp-content/themes/core/components/dialog/Dialog_Controller.php
```diff
@@ -0,0 +1,152 @@
...
+
+namespace Tribe\Project\Templates\Components\dialog;
+
+use Tribe\Libs\Utils\Markup_Utils;
+use Tribe\Project\Templates\Components\Abstract_Controller;
+use Tribe\Project\Templates\Components\Deferred_Component;
+use Tribe\Project\Templates\Components\container\Container_Controller;
+use Tribe\Project\Templates\Components\text\Text_Controller;
+
+class Dialog_Controller extends Abstract_Controller {
```
dpellenwood, line 12: Minor but there are some white space / alignment issues here that should be cleaned up before merging.

wp-content/themes/core/components/dialog/dialog.php
```diff
@@ -0,0 +1,43 @@
+<?php
+declare( strict_types=1 );
+
+use Tribe\Project\Templates\Components\dialog\Dialog_Controller;
+
+/*
```
dpellenwood, line 6: This is fantastic, Scott. Thanks for such good documentation here.

wp-content/themes/core/components/slider/css/slider.pcss
```diff
@@ -44,3 +74,87 @@
...
+		}
+	}
+}
+
+.c-slider__button--next,
+.swiper-container-rtl .c-slider__button--prev {
+
+	/* CASE: Dialog */
+	.c-dialog & {
+		right: -102px;
```
dpellenwood, line 141: Will this value work okay on small screens? It doesn't create horizontal scroll bars by chance?

dpellenwood (approved):
Some minor clean up and a couple questions. Other wise g2g for me! :squid: :dash:

PR #666, opened 2020-12-16
Update nginx.conf

nickpelton:
## What does this do/fix?

Adds a simple one liner doc update so folks know how to support multisite during setup using our docker stack.


jbrinley (approved):
huh, I thought a similar comment used to be here. Thanks.

PR #668, opened 2021-01-08
Feature/pass block id to theme

dpellenwood:
## What does this do/fix?
* Adds support for the [block editor's "Anchor" field](http://p.tri.be/0KQn0t) on our custom ACF blocks. This brings our blocks into parity with Core blocks in this respect.
* It also sets up the Base_Model, Block Models, & Block Controllers to better support other block properties, should we chose to utilize them in the future.
* Finally, it fixes an issue where some of our block controllers were using the `class_attribute()` utility method to return attributes. (They are all correctly using `concat_attrs()` now.)


ryanurban (approved):
Nice!

nickpelton (approved):
👍  For adding consistency. 

💭  Since all blocks need a get_attr() method and all use the same Markup_Utils class, would make sense to add this to a parent class or trait so in the future it's a one spot change.

dpellenwood:
>  Since all blocks need a get_attr() method and all use the same Markup_Utils class, would make sense to add this to a parent class or trait so in the future it's a one spot change.

Agreed. We also use the `get_classes()` method on all blocks. I'd like to consolidate both of those at the same time.

PR #672, opened 2021-01-27
remove travis badge for gha badge

nickpelton:
## What does this do/fix?

Travis tests were failing - because we don't use Travis anymore. Changed the build badge to GHA.


defunctl (approved):
Yay!

PR #673, opened 2021-02-02
Feature/post list plugin

msawicki:
## What does this do/fix?

Integrates the post list plugin. @ryanurban feel free to update anything here that needs to be. This should be integrated with the Post List and Card Grid.


ryanurban (approved):
AWESOME! Thanks so much @msawicki ❤️ 🌮

wp-content/plugins/core/src/Blocks/Types/Post_List/Post_List.php
```diff
@@ -10,35 +10,15 @@
 use Tribe\Libs\ACF\Repeater;
```
dpellenwood, line 10: I would really like to propose that we delete this Block entirely. It's not a "real" block as defined by the designs. It was build as a demo for the Post List field (Query) functionality. The purpose of this block is generally confusing for developers on projects.

dpellenwood (approved):
This is so wonderful, Mike. Thank you fro tackling it. Some day I will get a chance to poke at the JS. But for now, it's a huge improvement over our existing functionality.

dpellenwood:
@sc0ttkclark Any chance you would have the time to update the Content Loop Block to use The new PostList field after this merges? You can see the changes in the Card Grid here.

wp-content/plugins/core/src/Blocks/Types/Post_List/Post_List.php
ryanurban: Good call, agreed

ryanurban:
I just pushed the post list block removal. This should be ready to merge.

composer.json
```diff
@@ -81,6 +86,7 @@
     "mailgun/mailgun-php": "^2.8",
     "moderntribe/acf-image-select": "dev-master",
     "moderntribe/tribe-libs": "3.4.0",
+    "moderntribe/tribe-acf-post-list-field": "dev-main",
```
defunctl, line 89: @msawicki want to update this to "^1.0"?

defunctl (changes requested):

defunctl (approved):

ryanurban:
I'm gonna do one last thing here and update the remaining blocks now merged to use this new field before we merge 😎

PR #677, opened 2021-02-08
Fixes callback in Scroll To script

tjdunklee:
## What does this do/fix?

While working on another project, I noticed the callback for the `scroll-to.js` script had a typo in it. This fixes that callback, but also uses it to add auto element focusing to the `smooth-anchors.js` script. For accessibility reasons, we should change the focus for users if we are scrolling them to a certain position.

## Tests

Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests because this is optional functionality.


wp-content/themes/core/assets/js/src/utils/dom/smooth-anchors.js
```diff
@@ -19,6 +19,9 @@ const handleAnchorClick = ( e ) => {
 		offset: -150,
 		duration: 300,
 		$target: $( target ),
+		afterScroll: function() {
+			target.focus();
```
dpellenwood, line 23: This is great, except if the element being scrolled to isn't focusable. (such as adding an id attribute to an element which isn't a link or input, etc)  I [think this is a reasonable solution](https://github.com/moderntribe/codet-vision/blob/main/wp-content/themes/core/assets/js/src/utils/dom/smooth-anchors.js#L19-L27) to the issue. Happy to get other's input as well.

dpellenwood (approved):
Just one comment about focusing on the target. Thanks, Mike!

wp-content/themes/core/assets/js/src/utils/dom/scroll-to.js
```diff
@@ -33,7 +33,7 @@ const scrollTo = ( opts ) => {
 			}
 		}
 
-		$( 'html, body' ).animate( { scrollTop: position }, options.duration, options.easing, options.after_scroll );
+		$( 'html, body' ).animate( { scrollTop: position }, options.duration, options.easing, options.afterScroll );
```
ryanurban, line 36: Good catch ❤️

wp-content/themes/core/assets/js/src/utils/dom/smooth-anchors.js
ryanurban: Agreed, we should have this probably just be an additional bool argument you can opt into

ryanurban (approved):

tjdunklee:
Thank @ryanurban and @dpellenwood for the great suggestions. I have an idea that will solve both of them, I'll make an update soon and have you re-review.

tjdunklee:
I've added a boolean option called "auto_focus" that is off by default when you init the Smooth Anchors script. If it is set to TRUE, I took inspiration from Dave's Codet code to apply a tab-index to make any element focusable before setting the focus. Let me know what you think of this solution.

wp-content/themes/core/assets/js/src/utils/dom/smooth-anchors.js
```diff
@@ -19,20 +19,40 @@ const handleAnchorClick = ( e ) => {
 		offset: -150,
 		duration: 300,
 		$target: $( target ),
+		afterScroll: function() {
+			if ( ! options.auto_focus ) {
+				return;
+			}
+
+			if ( ! target.getAttribute( 'tabindex' ) ) {
+				target.setAttribute( 'tabindex', '-1' );
```
ryanurban, line 28: I just so happen to be digging into browser skip link issues so this matter is top of mind at the moment 😂

Fwiw, I was just reminded this utility only works on anchors that serve as an anchor hash link, so we don't need the tabindex functionality, and I'm going to commit the a11y tweaks I'm making to handle broken skip links, which I think will handle the focus bit.

Let's tag team this, I'll commit my work in this branch (still working on it, will have it wrapped today), you comment out and/or remove everything but the fixed callback, and then let's both test and see where we land on this? Thanks for noticing this and bringing it up!

tjdunklee: Sounds good, we can definitely tag team this.

I'm a little confused though why you wouldn't need the tabindexing. The smooth scroll's destination could be anything with an `id` so like Dave mentioned it could be a non-focusable element. I'm picturing an author linking in a nav menu item to a specific block on that page.

Either way, I'll wait for your commit and make tweaks to mine based on that.

ryanurban: Scope this: https://github.com/moderntribe/square-one/pull/677/files#diff-3322c350466fd62a27abec4bb2f60e0638b414251d64b083108843ccf413c699R37

We only apply this overall utility to anchors that have an anchor hash href value : )

ryanurban: Oh derp, ignore me, right, the thing it might reference may not be focusable. Haha, what I'm doing up for a11y will also address that too : )

tjdunklee: No problem at all! Great! Looking forward to seeing what you come up with.

ryanurban:
@tjdunklee scope this and let me know what you think and if it makes sense. I do wonder if we should keep/have the auto focus option for the smooth scroll functionality? Would there ever be a case where we wouldn't want to add focus to the targeted element? My gut, for a11y sake, says no. What do you think?

ryanurban:
My additional fixes the browser issue where skip links don't work correctly. As in you click on the skip link (or any anchor hash for that matter), it positions you to the correct placement on the page, but doesn't shift focus as it should to the element that you've skipped to and you wind up very annoyingly tabbing back up at the top of the site.

This fixes it across the board and handles adding/removing of tabindex for the targeted element based on whether or not it's a focusable item. I also added a custom event to ensure this works when smooth scroll is active on a project as due to how we handle that (preventing the default behavior and then adding the hash with pushstate), the `hashchange` event doesn't get triggered correctly otherwise. I figured it'd be nice to handle the focus management in one place.

tjdunklee:
Everything is looking good to me @ryanurban! I gave it a quick test locally visually and also with VoiceOver. I like it. Do you want to update the changelog before we merge in?

PR #679, opened 2021-02-15
Blocks/sqone 512/lead form refinements

ScottGarrison:
## What does this do/fix?

Adds some additional refinements to the current Lead Form Block
- Background color
- Form position
- Form Field position

image


wp-content/themes/core/integrations/gravity-forms/css/default.pcss
```diff
@@ -23,6 +23,33 @@
...
+		/* CASE: Lead Form: Background Dark */
+		.b-lead-form--dark & {
+			color: var(--color-white);
+		}
+	}
+
+	.gform_footer {
+		/* CASE: Lead Form: Form Fields Inline */
+		.b-lead-form--fields-inline & {
+			margin-left: var(--spacer-10);
```
ryanurban, line 51: With these styles being more about the block than the integration itself, I'm wondering if it would be more helpful to add these styles within the block css vs. here to keep the separation between block and integration cleaner?

wp-content/themes/core/integrations/gravity-forms/css/spinner.pcss
```diff
@@ -21,6 +21,12 @@
 	@media (--viewport-medium) {
 		padding-right: 50px;
 	}
+
+	/* Case: Center Button so need to remove right padding */
+	.b-lead-form--layout-bottom & {
+		padding-right: 0;
+		text-align: center;
+	}
```
ryanurban, line 29: Same question here.

ryanurban (commented):
Thanks, Scott!

wp-content/themes/core/integrations/gravity-forms/css/default.pcss
ScottGarrison: Previous block notes kinda had this type of setup where if you're modifying a class, make it a case and put it within that file. I could see situations where other blocks need an inline form or a white label on a black background. Maybe keep them here but make the class names more reusable like `.is-form-inline` ?

ryanurban: Noting our follow up to dos:

- Will update the color/theme based classed to leverage existing SQ1 `t-theme--light` for consistency
- The inline layout helper class is really contextually centered on the GF integration and could be handy to have available outside of the context of the lead form block or any block for that matter. We'll turn it into a complex layout utility that is GF only

Thanks Scott!

wp-content/themes/core/components/blocks/lead_form/css/lead-form.pcss
```diff
@@ -86,26 +82,20 @@
...
 	margin-top: var(--spacer-30);
+	max-width: var(--grid-5-col);
 
-	/* CASE: Lead Form: Grid Width */
-	.b-lead-form--width-grid & {
-		@media (--viewport-full) {
-			margin-top: var(--spacer-40);
-		}
+	@media (--viewport-full) {
+		width: calc(41.6667% - var(--grid-gutter-half));
```
dpellenwood, line 88: Minor, but it might be nice to add a short comment here mentioning the number of columns + gutter this equates to. I.e - something like `/* 5 of 12 columns + gutter */`

wp-content/themes/core/integrations/gravity-forms/css/complex-layouts.pcss
```diff
@@ -76,3 +76,25 @@
...
 	}
 }
+
+/*
+* Utility class to use for inline forms.
+* Set up only for an individual form field and Submit button
+*/
+
+.gform_inline {
+	.gform_wrapper {
```
dpellenwood, line 86: Noice!

dpellenwood (approved):
One super minor comment note. Looks great to me though. Thanks, Scott!

wp-content/themes/core/integrations/gravity-forms/css/label.pcss
```diff
@@ -4,6 +4,11 @@
 
 .gfield_label {
 	@mixin form-label;
+
+	/* CASE: Theme Light */
+	.t-theme--light & {
+		color: var(--color-white);
+	}
```
ryanurban, line 11: Similar to what we do in other typographic mixins, we could potentially just add this to the `form-label` mixin so it'd work everywhere its used?

ryanurban (approved):
❤️

PR #680, opened 2021-03-03
fixing the Gallery grid default to an array instead of a string

ScottGarrison:
## What does this do/fix?

Updates an incorrect default from string to array


ryanurban (approved):
Thanks Scott!

PR #681, opened 2021-03-04
Address slider component issues

ryanurban:
## What does this do/fix?

* Updates the swiper js library to latest version
* Addresses issues that were breaking the slider component (incomplete port on my part last year)
* Updated gulp postcss config to preserve custom properties by default and to allow for native use of :focus-within & :focus-visible

## Tests

Does this have tests?

- [ ] Yes

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,10 @@
 
 All notable changes to this project will be documented in this file.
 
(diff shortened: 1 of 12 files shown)


ScottGarrison (approved):
:fire:

PR #683, opened 2021-03-05
Updates linting to include JS and CSS files.

mmcachran:


ryanurban:
@mmcachran is this for the PR linting or the existing build system?

mmcachran:
> @mmcachran is this for the PR linting or the existing build system?

This is for any staged files getting committed to the repo. PHP linting is already in place, this just adds it for JS and CSS.

ryanurban:
Ah ok cool. Only reason I ask is we already have eslint & stylelint also setup for the theme itself and our build system, so wanted to make sure we weren't duplicating and doubling up 👍 Thanks for doing this ⚡

jamiepaul:
@mmcachran does this work w/ the husky linting messages pre-commit? To @ryanurban's point, just want to make sure we're not doubling up.

mmcachran:
> @mmcachran does this work w/ the husky linting messages pre-commit? To @ryanurban's point, just want to make sure we're not doubling up.

Yes, this is for pre-commit linting with Husky.

ryanurban:
Hmm, makes me wonder if this is duplicative then, though I do like the fact that it's automated. Would you be able to leverage our existing `gulp lint` task @mmcachran and hook it up to be run as part of this instead?

PR #684, opened 2021-03-22
Update lead form block GF implementation

ryanurban:
## What does this do/fix?

This fixes a bug Belinda noted, where the scripts for Gravity Forms were not being enqueued for our programmatic implementation of GF in the lead form block. I went ahead and added the ability for this block to support inner blocks, but only the GF block. This does add a few wrappers in the preview, none in the FE, and everything looks ok visually, so going to roll with it (dipping our toes into the inner block pool, it's gonna be great). Either way we'd want to have the ability to support a given GF in this block with the various options available vs. having them be fixed in the template, and this is a solid use case for using the inner block vs. recreating the GF block here. Inner blocks FTW! 😂 

## Tests

Does this have tests?

- [ ] Yes


verytwisty (approved):
I didn't know you could do innerBlocks with PHP! Cool thank you!

wp-content/themes/core/components/blocks/lead_form/Lead_Form_Block_Controller.php
```diff
@@ -255,6 +244,8 @@ public function get_form(): string {
 			return '';
 		}
 
-		return gravity_form( $this->form, false, false, false, null, false, 1, false );
+		$allowed_blocks = [ 'gravityforms/form' ];
+
+		return '<InnerBlocks allowedBlocks="' . esc_attr( wp_json_encode( $allowed_blocks ) ) . '" />';
```
nickpelton, line 249: NIT: Typically like to use sprintf for in code string management.

wp-content/themes/core/components/blocks/lead_form/Lead_Form_Block_Controller.php
```diff
@@ -255,6 +244,8 @@ public function get_form(): string {
 			return '';
 		}
 
-		return gravity_form( $this->form, false, false, false, null, false, 1, false );
+		$allowed_blocks = [ 'gravityforms/form' ];
```
nickpelton, line 247: Should this be a method and or filterable JIC we need to add additional blocks?

wp-content/themes/core/components/document/header/header.php
```diff
@@ -25,7 +25,7 @@
 	<?php // PLATFORM META: IE ?>
 	<meta name="application-name" content="<?php bloginfo( 'name' ); ?>">
 
-	<?php do_action( 'wp_head' ) ?>
+	<?php do_action( 'wp_head' ); ?>
```
nickpelton, line 28: 🚀

nickpelton (approved):

wp-content/themes/core/components/blocks/lead_form/Lead_Form_Block_Controller.php
ryanurban: Great question, at some point yes, I think that's correct. For now or at least until we need it, I think this is ok as we only tend to use GF.

ryanurban: Noted as a future TODO

ryanurban:
Belinda had a great suggestion. Using a block template here instead of just an inner block. That allows us to lock this block to only use this particular, required / necessary inner block configuration, and to set it as the default as well.

ScottGarrison (approved):

PR #693, opened 2021-04-02
Remove old backend doc links

lindseyolson:
## What does this do/fix?
Removes links from the Backend Readme that no longer exist. I believe all documentation is now just in the `/docs` folder. I also removed 2 of the plugins that I believe are no longer being used.

## Tests
Does this have tests?
- [ ] Yes
- [X] No, this doesn't need tests because it's just a documentation update.


nickpelton (approved):
Awesome! Ship it! 🚀

PR #706, opened 2021-04-26
Static Analysis With Psalm

tarecord:
## What does this do/fix?
Adds [Psalm](https://github.com/vimeo/psalm/) to handle static analysis of the codebase. Right now, it's only consuming and reviewing `wp-content/plugins/core/src`. It's also pulling in a [Human Made plugin](https://github.com/humanmade/psalm-plugin-wordpress) that configures Psalm to use WordPress stubs so it doesn't throw errors for missing functions/types.

## Todo
- [ ] ~Clean up current errors with the core plugin to start with a "clean slate" with new development and avoid failing builds~ Psalm provides a way to set up a baseline file to ignore current errors—if you don't have time to fix them all.
- [x] Set up Github Actions to automatically run Psalm checks


nickpelton:
Neat! 

I think we should also add the path for the theme and cover all our generated code. I would also make sure we address all the PHP errors we'll see before adding the GIthub action otherwise we'll be getting failed builds. Are you planning on addressing the code issues in this branch? Or another one?

tarecord:
I totally agree, I wanted to get a PR up to start the discussion.

I was thinking about handling the current errors in another branch for now, and merging that into this branch so we can keep the commit history clean.

nickpelton:
Sounds like a good appoach!

defunctl:
Love this! We should only commit a `psalm.xml.dist` and put the `pslam.xml` in .gitignore to allow developers to customize/change it locally if needed.

Would this be something that could be integrated into husky hooks library we have to automatically run?

tarecord:
Psalm provides a way to ignore all the current errors in the codebase so you don't have to spend a massive amount of time fixing all the issues upfront. They can be fixed moving forward and we can begin using Psalm with new code:
>If you have a bunch of errors and you don't want to fix them all at once, Psalm can grandfather-in errors in existing code, while ensuring that new code doesn't have those same sorts of errors.
```
vendor/bin/psalm --set-baseline=your-baseline.xml
```
>will generate a file containing the current errors. You can commit that generated file so that Psalm running in other places (e.g. CI) won't complain about those errors either, and you can update that baseline file (to remove references to things that have been fixed) with
```
vendor/bin/psalm --update-baseline
```
>Baseline files are a great way to gradually improve a codebase.

wp-content/plugins/core/src/Admin/Admin_Subscriber.php
```diff
@@ -21,29 +21,29 @@ private function editor(): void {
...
-			return $this->container->get( Editor_Styles::class )->enqueue_mce_editor_styles();
+		add_action( 'admin_init', function (): void {
+			$this->container->get( Editor_Styles::class )->enqueue_mce_editor_styles();
 		}, 10, 1 );
 	}
 
-	private function editor_formats() {
-		add_filter( 'teeny_mce_buttons', function ( $buttons, $editor_id ) {
+	private function editor_formats(): void {
+		add_filter( 'teeny_mce_buttons', function ( array $buttons, string $editor_id ): array {
```
defunctl, line 40: Is there any way to make it ignore the args in the closure, these should not be typed because WordPress does not send reliable types.

It should be like this:

```php
add_filter( 'teeny_mce_buttons', function ( $buttons, $editor_id ): array {
			return $this->container->get(Classic_Editor_Formats::class)->teeny_mce_buttons( (array) $buttons, (string) $editor_id );
		}, 10, 2 );
```

wp-content/plugins/core/src/Blocks/Types/Buttons/Buttons.php
```diff
@@ -21,6 +21,9 @@ class Buttons extends Block_Config {
 	public const STYLE_SECONDARY = 'secondary';
 	public const STYLE_CTA       = 'cta';
 
+	/**
+	 * @return void
+	 */
 	public function add_block() {
```
defunctl, line 27: Weird it didn't add `: void` here and below

tarecord: It looks at PHPDoc tags to detect types, but I wonder if there's a way to force it to add the `: type` syntax. https://psalm.dev/docs/annotating_code/supported_annotations/

tarecord: Seems like we only need this to force the syntax.
```xml
<psalm
  useDocblockTypes="false"
>
```

wp-content/plugins/core/src/Admin/Admin_Subscriber.php
tarecord: we _could_ [globally ignore closure return types](https://psalm.dev/docs/running_psalm/issues/MissingClosureReturnType/) with `MissingClosureReturnType` in the [config suppression](https://psalm.dev/docs/running_psalm/dealing_with_code_issues/#config-suppression), but I'm sure we use them elsewhere we'd like to check types.

defunctl: Ahh that rule is for the return type of the closure, not the parameters being passed into it.

wp-content/plugins/core/src/Blocks/Types/Buttons/Buttons.php
defunctl: Nice, the new PHPCS rules enforce this as well, so might not be needed?

nickpelton:
Giving this a test today and have a few questions. PHPCS also does a lot of these checks. How do we avoid doubling up and maintaining two possibly conflicting rule sets?

**Case**: Currently after  merging main, the commit hook PHPcs check won't allow commits after conflicting Psalm changes.

```Found existing alias for "git commit -m". You should use: "gcmsg"
⠴ waiting ^R                                                                                                                                                                                           
  EXECUTE > phpcs-core-plugin
 ...............................E...E.............E.......... 60 / 65 (92%)
.....                                                        65 / 65 (100%)



FILE: wp-content/plugins/core/src/Admin/Admin_Subscriber.php
----------------------------------------------------------------------
FOUND 2 ERRORS AFFECTING 1 LINE
----------------------------------------------------------------------
 42 | ERROR | [x] Expected 1 spaces after opening parenthesis; 0
    |       |     found
    |       |     (PEAR.Functions.FunctionCallSignature.SpaceAfterOpenBracket)
 42 | ERROR | [x] Expected 1 spaces before closing parenthesis; 0
    |       |     found
    |       |     (PEAR.Functions.FunctionCallSignature.SpaceBeforeCloseBracket)
----------------------------------------------------------------------
PHPCBF CAN FIX THE 2 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------


FILE: ...plugins/core/src/Integrations/Google_Tag_Manager/GTM_Scripts.php
----------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
 57 | ERROR | [x] Expected 1 line between description and
    |       |     annotations, found 0.
    |       |     (SlevomatCodingStandard.Commenting.DocCommentSpacing.IncorrectLinesCountBetweenDescriptionAndAnnotations)
----------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------


FILE: wp-content/plugins/core/src/Shortcodes/Gallery.php
----------------------------------------------------------------------
FOUND 2 ERRORS AFFECTING 2 LINES
----------------------------------------------------------------------
 108 | ERROR | [x] Referencing general \Exception; use \Throwable
     |       |     instead.
     |       |     (SlevomatCodingStandard.Exceptions.ReferenceThrowableOnly.ReferencedGeneralException)
 121 | ERROR | [x] There must be one empty line before class closing
     |       |     brace.
     |       |     (SlevomatCodingStandard.Classes.EmptyLinesAroundClassBraces.NoEmptyLineBeforeClosingBrace)
----------------------------------------------------------------------
PHPCBF CAN FIX THE 2 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------

Time: 3.64 secs; Memory: 12MB



SUMMARY: (done in 3.93 seconds)
✔️  phpcs-mu
✔️  phpcs-core-theme
🥊  phpcs-core-plugin
```

nickpelton (changes requested):
Incompatible with existing PHPcs rules.

nickpelton:
Another note. The GitHub actions Workflow name matches the existing `PHP Composer and Codesniffer` action. We should rename it if this is a separate analysis task. OR we should put them together since both require a build step and we can save CI minutes.

.github/workflows/static-analysis.yml
```diff
@@ -0,0 +1,59 @@
...
+
+    env:
+      cache_version: 0 # increment to flush all caches
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Add GitHub oAuth
+        env:
+          TOKEN: ${{ secrets.GH_TOKEN }}
```
defunctl, line 27:
```suggestion
          TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

composer.json
```diff
@@ -132,6 +133,7 @@
     "phpunit/phpunit": "^8.0 || ^9.0 <9.5",
     "sirbrillig/phpcs-variable-analysis": "^2.0.1",
     "squizlabs/php_codesniffer": "^3.4.2",
+    "vimeo/psalm": "^4.7",
```
defunctl, line 136: What's this doing in here?

wp-content/plugins/core/src/Blocks/Types/Card_Grid/Card_Grid.php
```diff
@@ -29,7 +29,9 @@ class Card_Grid extends Block_Config {
 	public const LAYOUT_INLINE    = 'inline';
 
 	/**
-	 * Register the block
+	 * 	 * Register the block
```
defunctl, line 32: Not sure what's up with spacing on these comments or if I did that in another PR by accident?

wp-content/plugins/core/src/Blocks/Types/Hero/Hero.php
```diff
@@ -25,7 +25,9 @@ class Hero extends Block_Config {
 	public const LAYOUT_CENTER    = 'center';
 
 	/**
-	 * Register the block
+	 * 	 * Register the block
```
defunctl, line 28: Yeah, not sure what's going on here?

wp-content/plugins/core/src/Blocks/Types/Interstitial/Interstitial.php
```diff
@@ -22,6 +22,9 @@ class Interstitial extends Block_Config {
 	public const LAYOUT_LEFT      = 'left';
 	public const LAYOUT_CENTER    = 'center';
 
+	/**
+	 * @return void
+	 */
 	public function add_block() {
```
defunctl, line 28: Would be way better to just add `add_block(): void {` to all of these remove the @return docBlock. We don't need return comments if the method specifies what they are unless it's an array and then we should specify what type of array it is.

wp-content/plugins/core/src/Cache/Listener.php
```diff
@@ -21,27 +24,35 @@ public function save_post( $post_id, $post ) {
...
-	 * A p2p relation has been created.
+	 * 	 * A p2p relation has been created.
+	 * 	 *
 	 *
 	 * @param $p2p_id
+	 *
+	 * @return void
 	 */
-	public function p2p_created_connection($p2p_id) {
+	public function p2p_created_connection($p2p_id): void {
```
defunctl, line 34:
```suggestion
	public function p2p_created_connection( $p2p_id ): void {
```

wp-content/plugins/core/src/Integrations/Google_Tag_Manager/GTM_Scripts.php
```diff
@@ -16,9 +16,12 @@ public function __construct( Analytics_Settings $settings ) {
 	}
 
 	/**
-	 *  Google Tag Manager (head tag)
+	 * 	 *  Google Tag Manager (head tag)
```
defunctl, line 19: Is it psalm that's mangling all of these docblocks?

defunctl (changes requested):
Can we configure this to only specify @return docblocks when it's an array or we are not returning a specific type defined method type hints?

dpellenwood (commented):
This is a bit out of my wheel house, so I'm going to defer to others on this PR.

composer.json
tarecord: not sure what you mean by this?

nickpelton:
Hey @tarecord , I still this this would be a valuable addition. Do you think you'll have time to work on this in the near future?

tarecord:
Closing in favor of #971

PR #708, opened 2021-04-29
initial buildout of spacer block

ScottGarrison:
## What does this do/fix?

*TICKET* - https://moderntribe.atlassian.net/browse/SQONE-541

This adds a new Spacer block to the base SQ1 system.
Gives the ability for content creators to add additional space above and blow other blocks.
In the case of sites that have block color options, this would give a way for content creators to expand the spacing of a blocks "padding" above and below.
This version gives gives an admin view of a dashed border and a "Spacer" text to help content creators see where the spacer block is.

image


wp-content/plugins/core/src/Blocks/Types/Spacer/Spacer_Model.php
```diff
@@ -0,0 +1,19 @@
+<?php
```
ryanurban, line 1: It's probably still worth this block supporting the custom classes and definitely the anchor capability like the other blocks and core blocks do.

wp-content/themes/core/components/blocks/spacer/block-editor.pcss
```diff
@@ -0,0 +1,22 @@
...
+ * Adding admin styles for the Spacer Block.
+ *
+ * ----------------------------------------------------------------------------- */
+
+.wp-block[data-type^="acf/spacer"] {
+	.b-spacer {
+		border: 1px dashed #c8c8c8;
+
+		&:before {
+			content: '[SPACER]';
```
ryanurban, line 15: It might be worth getting this out of the CSS and having it be part of the markup that is only visible in the editor and then display none'd in the FE? Thinking about the need for supporting other languages primarily.

ryanurban (commented):
Nice. I could see this becoming more robust in the future, but good first step (allowing for custom spacing, custom spacing for mobile vs. desktop, more breakpoints, ability to choose a separator graphic, etc.).

wp-content/plugins/core/src/Blocks/Types/Spacer/Spacer_Model.php
ryanurban: Another thought for the future, but we will need to do some thinking around anchor links and how that plays out for focus/a11y if we plan to use a block to provide spacing between blocks. And within that convo there are some fun gotchas I experienced on Harvard when using bodylock, so we should chat about that at a future date and I can share a couple of solutions.

nickpelton (approved):

wp-content/themes/core/components/blocks/spacer/block-editor.pcss
ScottGarrison: I'm not sure on this. I realize that it's text that should be translated but it's sole purpose is visualization so that editors know that the block even exists in that space. Having extra markup on the front end just for that seems unnecessary. 

Maybe instead of copy it should just be something like [------------] , or even an emoji :space_invader:

ChrisMKindred: Why not mimic the UX that already exists in the core spacer block and give it a background color? http://p.tri.be/RuMOsc

ryanurban: For sure, totally get that. Thinking of like WP admin also being in other languages. Probably a vote for mimicking core and/or removing the text piece as you noted.

wp-content/themes/core/assets/css/src/theme/layouts/default.pcss
```diff
@@ -71,7 +71,7 @@ main {
 
 .l-sink {
 
-	& > *:not(.c-block):not(.alignwide):not(.alignfull) {
+	& > *:not(.c-block):not(.alignwide):not(.alignfull):not(.b-spacer) {
```
dpellenwood, line 74: I worry about adding additional conditions here as it's already more gnarly than it should be. Would it be possible to just use one the the existing classes (`alignwide` or `alignfull`, maybe) to accommodate this? (Here and on line 101)

wp-content/themes/core/components/blocks/spacer/Spacer_Block_Controller.php
```diff
@@ -0,0 +1,80 @@
...
+		return [
+			self::CLASSES => [ 'b-spacer' ],
+		];
+	}
+
+	/**
+	 * @return string
+	 */
+	public function get_attrs(): string {
+		return Markup_Utils::class_attribute( $this->attrs );
```
dpellenwood, line 63: `concat_attrs()`, not `class_attribute()`. ;)

wp-content/themes/core/components/blocks/spacer/css/spacer.pcss
```diff
@@ -0,0 +1,40 @@
...
+		height: 48px;
+	}
+
+	/* CASE: Large Size */
+	&.b-spacer--large {
+		height: 48px;
+
+		@media (--viewport-medium) {
+			height: 96px;
+		}
```
dpellenwood, line 20: Can we use existing spacer vars rather than arbitrary px values?

dpellenwood (approved):
Some small requests, but looks good overall. Thanks Scott!

wp-content/themes/core/components/blocks/spacer/css/spacer.pcss
ScottGarrison: My concern was that design would update the spacer in the kitchen sink and effect this spacer block. If that's expected than thats fine but design seemed pretty specific on this set of pixels

wp-content/themes/core/assets/css/src/theme/layouts/default.pcss
ScottGarrison: Good call. I will give that a go

wp-content/themes/core/components/blocks/spacer/css/spacer.pcss
dpellenwood: I think that would be an intentional change then, yes. It would affect any spacing throughout the system, including the spacing here/between blocks.

wp-content/themes/core/components/blocks/spacer/block-editor.pcss
ScottGarrison: So I tested core spacer and the issue is that the background color goes away after focus is removed. The issue the content editors were having was they were losing track of the block and couldn't figure out if there was actually anything there. Mimicking in this case doesn't solve that.

ChrisMKindred: Does the [List View](http://p.tri.be/v7Ojfx) or changing to [Select Mode](http://p.tri.be/DCQBIA) solve that issue? 

It seems like this is a training issue. It would cause more confusion to show something that isn't going to display on the FE causing users to try to remove the extra design items because they don't realize that this doesn't work the same as other blocks.

ScottGarrison: While I agree with you, I removed the text from outputting and now there's just a top and bottom dashed border. I'm going to leave the rest up to the the design team since this was their scope

PR #710, opened 2021-04-30
Miscellaneous maintenance

ryanurban:
## What does this do/fix?

This updates a handful of items:

- Updated: refreshed local-config-sample.php to have a more thorough set of constants you may need, especially now that we have a Local environment option.
- Updated: newer version of spin.js requires some css, added that.
- Updated: set a default set of theme image paths for postcss-inline-svg
- Updated: disabled cssnano discarding "unused" keyframes. It was borking the spin.js css.
- Updated: remove the tribe/body_opening_tag action in favor of wp_body_open and updated our GTM implementation to use it.
- Fixed: body-lock.js needed an update to work correctly for safari/ios.
- Updated: added a isLocked check for body-lock.js, it's handy. 

## QA

Links to relevant issues
- [SQONE-560](https://moderntribe.atlassian.net/browse/SQONE-560)
- [SQONE-132](https://moderntribe.atlassian.net/browse/SQONE-132)

## Tests

Does this have tests?

- [ ] Yes

### DIFF
--- a/local-config-sample.php
+++ b/local-config-sample.php
@@ -7,8 +7,38 @@
  *
  * The default WordPress databases defines are automatically populated in
  * dev/docker/docker-compose.yml which you can override here with custom
- * defines e.g. define( 'DB_NAME', 'tribe_square1' );
-*/
+ * defines, e.g. define( 'DB_NAME', 'tribe_square1' ). If you are using Local,
+ * you will need to define these here.
+ */
+//define( 'DB_NAME', 'tribe_square1' );
+//define( 'DB_USER', 'root' );
+//define( 'DB_PASSWORD', 'root' );
+//define( 'DB_HOST', 'localhost' );
+//define( 'DB_TABLE_PREFIX', 'wp_' );
+
+/*
+ * Multisite
+ *
+ * If you enable multisite in wp-config.php, ensure to provide your local.tribe URI here
+ */
+//define( 'DOMAIN_CURRENT_SITE', 'square1.tribe' );
+
+/*
+ * Debugging
+ */
+define( 'WP_DEBUG', true );
+define( 'WP_DEBUG_LOG', true );
+define( 'WP_DEBUG_DISPLAY', true );
+define( 'SAVEQUERIES', true );
+define( 'SCRIPT_DEBUG', true );
+define( 'WP_CACHE', false );
+
+/*
+ * Panels
+ *
+ * If your project uses panels this will disable/enable panels caching.
+ */
+//define( 'TRIBE_DISABLE_PANELS_CACHE', true );
 
 /*
  * React dev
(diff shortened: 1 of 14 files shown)


wp-content/plugins/core/src/Integrations/Google_Tag_Manager/Google_Tag_Manager_Subscriber.php
```diff
@@ -10,7 +10,7 @@ public function register(): void {
 		add_action( 'wp_head', function () {
 			$this->container->get( GTM_Scripts::class )->inject_google_tag_manager_head_tag();
 		} );
-		add_action( 'tribe/body_opening_tag', function () {
+		add_action( 'wp_body_open', function () {
```
ChrisMKindred, line 13: 🥇 This will make other 3rd party plugins happy.

local-config-sample.php
```diff
@@ -7,23 +7,63 @@
...
+
+/*
+ * Debugging
+ */
+define( 'WP_DEBUG', true );
+define( 'WP_DEBUG_LOG', true );
+define( 'WP_DEBUG_DISPLAY', true );
+define( 'SAVEQUERIES', true );
+define( 'SCRIPT_DEBUG', true );
+define( 'WP_CACHE', false );
```
ChrisMKindred, line 34: Would it be cleaner to set the [environment variable](https://make.wordpress.org/core/2020/08/27/wordpress-environment-types/) to `local` and let [wp-config-environment.php](https://github.com/moderntribe/square-one/blob/e410ce26ba03796d914896e65564d5dab6c0009d/wp-config-environment.php#L158) handle these in the config_defaults?

ChrisMKindred (approved):
Looks good @ryanurban. Just one real question but it isn't a blocker.

dpellenwood (approved):
Looks great! Thanks for grabbing these.

wp-content/plugins/core/src/Integrations/Google_Tag_Manager/Google_Tag_Manager_Subscriber.php
ryanurban: Tip 'o the cap to Nick ❤️

local-config-sample.php
nickpelton: I agree. We have some deployments where we us the wp-config-local as a template for other environments, so I'd rather key off `WP_ENVIRONMENT_TYPE`.

nickpelton (changes requested):
Small change requested.

local-config-sample.php
tarecord: I'll just leave this here... https://github.com/moderntribe/square-one/pull/703#issuecomment-827605757
```php
/**
 * Set the current environment type. Accepted values:
 * - production (default)
 * - staging
 * - development
 * - local
 *
 * @link https://make.wordpress.org/core/2020/07/24/new-wp_get_environment_type-function-in-wordpress-5-5/
 */
define( 'WP_ENVIRONMENT_TYPE', 'local' );

// Fallback for older environments.
define ( 'ENVIRONMENT', 'LOCAL' );
```

nickpelton: Perfect 😂

ryanurban:
@ChrisMKindred @nickpelton @tarecord give that a final look see?

nickpelton (approved):

ChrisMKindred (approved):
👍

PR #715, opened 2021-05-13
Test driving routes. Documentation. Potential iterations.

nickpelton:
## What does this do/fix?

* Tests the Routes feature to see how usable it is. Note some potential future improvements.
* Moved the entrypoint feature flag (via comment) to the subscriber for consistency
* Added some additional documentation for variations of routes.
* ~Tested and logged an issue:  https://github.com/moderntribe/square-one/issues/716~

### DIFF
--- a/wp-content/plugins/core/src/Core.php
+++ b/wp-content/plugins/core/src/Core.php
@@ -61,7 +61,6 @@ class Core {
 		Google_Tag_Manager_Subscriber::class,
 		Gravity_Forms_Subscriber::class,
 		Nav_Menus_Subscriber::class,
-		Routes_Subscriber::class,
 		Shortcodes_Subscriber::class,
 		Theme_Subscriber::class,
 		Yoast_SEO_Subscriber::class,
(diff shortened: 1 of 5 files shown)


nickpelton:
@mmcachran I had some trouble using the routes query_vars. Documented the issue [here](https://github.com/moderntribe/square-one/issues/716). Let me know when you have a chance to review with me.

nickpelton:
Chatted with Matt and some good takeaways:

* Cache is really a permalinks flush. Need to make this clearer it's not really a cache manger, more a schema trigger.
* Permalink flush version is tracked at the subscriber level, but maybe should be tracked at a route level.Each route could have a version and trigger a flush if needed. Avoids the mistake of forgetting to change a version number somewhere else.
* The definers are initializing the routes. Compared to other patterns, these should be moved to the main subscriber it's more obvious what routes are active -similar to other subscriber patterns.

nickpelton:
Fixed in last sprint release.

PR #717, opened 2021-05-14
[SQONE-566]: Global Block Field injection

defunctl:
https://moderntribe.atlassian.net/browse/SQONE-566

## What does this do/fix?

WIP

Adds the ability to inject global field data into blocks and their controllers. Handy for adding global block fields that would set attributes/classes on block controllers.

Companion Tribe Libs PR: https://github.com/moderntribe/tribe-libs/pull/83

## Tests

Does this have tests?

- [X] Yes, it will.


wp-content/themes/core/components/blocks/card_grid/Card_Grid_Controller.php
```diff
@@ -76,7 +76,7 @@ protected function defaults(): array {
 	protected function required(): array {
 		return [
 			self::CONTAINER_CLASSES => [ 'b-card-grid__container', 'l-container' ],
-			self::LOOP_CLASSES      => [ 'b-card-grid__loop' ],
+			self::LOOP_CLASSES      => [ 'b-card-grid__loop', 't-theme--light' ],
```
dpellenwood, line 79: @ScottGarrison, Curious why the t-theme--light class is necessary, especially as a required class. Is this the default anyway?

wp-content/themes/core/components/card/css/card.pcss
```diff
@@ -157,6 +159,12 @@
 	@mixin t-body-small;
 
 	color: var(--color-grey);
+
+	/* CASE: Dark Theme */
+	.t-theme--dark & {
+		color: var(--color-white);
+		opacity: 0.8;
```
dpellenwood, line 166: @ScottGarrison Why is the whole card's opacity set here? Wouldn't we only want to adjust the background color opacity?

dpellenwood (commented):
Couple small questions for @ScottGarrison when he returns. These are ported from #718

wp-content/themes/core/components/blocks/card_grid/Card_Grid_Controller.php
ScottGarrison: because of the hierarchy. If the block itself gets `t-theme--dark`, then the card itself gets a black background with white text. It looked a little strange that way so I just override it here so the block itself gets the black background but the cards are the normal white background with black text. It might not need to be part of the "required" classes

wp-content/themes/core/components/card/css/card.pcss
ScottGarrison: Looks like this isn't the whole card, it's just the card date. The original designs have an off grey color for the date instead of straight black. This was a bit of a workaround to keep that styling the same.

defunctl:
Closing this as the backend is now implemented with Block Middleware

PR #651, opened 2020-10-23
Replace Safe SVG

jbrinley:
## What does this do/fix?

Removes the Safe SVG plugin, replacing it with the update from https://github.com/moderntribe/tribe-libs/pull/71

## Tests

Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests because... it is a dependency update
- [ ] No, I need help figuring out how to write the tests.

PR #721, opened 2021-05-17
Misc Bug fixes & clean up

dpellenwood:
## What does this do/fix?
* Fixes an issue related to public paths for JS module resolution. (The paths used for resolution were duplicating the base path like so: `/wp-content/themes/core/wp-content/themes/core/...`.)
* Drop the Composer plugin requirement for Safe SVG. Improved functionality is baked into SquareOne already. See PR #651 & related moderntribe/tribe-libs#71
* Updates the Limit Login Attempts force-activation path per the updated version of the plugin.

### DIFF
--- a/webpack/admindev.js
+++ b/webpack/admindev.js
@@ -21,7 +21,7 @@ module.exports = merge.strategy( {
 	externals,
 	output: {
 		path: resolve( `${ __dirname }/../`, pkg.square1.paths.core_admin_js_dist ),
-		publicPath: `/wp-content/themes/core/${ pkg.square1.paths.core_admin_js_dist }`,
+		publicPath: `/${ pkg.square1.paths.core_admin_js_dist }`,
 	},
 	plugins: [
 		new MiniCssExtractPlugin( {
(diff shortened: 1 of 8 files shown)


nickpelton (approved):

webpack/admindev.js
```diff
@@ -21,7 +21,7 @@ module.exports = merge.strategy( {
 	externals,
 	output: {
 		path: resolve( `${ __dirname }/../`, pkg.square1.paths.core_admin_js_dist ),
-		publicPath: `/wp-content/themes/core/${ pkg.square1.paths.core_admin_js_dist }`,
+		publicPath: `/${ pkg.square1.paths.core_admin_js_dist }`,
```
faction23, line 24: Godd catch. I gapped when updating webpack here. I think you can just remove publicPath now altogether https://github.com/webpack/webpack/issues/11127 I did on my other projects and it works great.

faction23 (approved):
Thanks for the catch on the chunk paths!

webpack/admindev.js
dpellenwood: Ooo, Nice! I'll give that a shot here. Thank you!

PR #727, opened 2021-05-20
[SQONE-543] Modify text component to check for null

lindseyolson:
## What does this do/fix?
This PR updates the text component to check for null, instead of empty. In this case, it will fix the issue of a "0" being considered not valid content.

## Tests
Does this have tests?
- [ ] Yes
- [X] No, this doesn't need tests because it's a bug fix


wp-content/themes/core/components/text/text.php
```diff
@@ -9,7 +9,7 @@
 // phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UndefinedVariable
 $c = Text_Controller::factory( $args );
 
-if ( empty( $c->get_content() ) ) {
+if ( $c->get_content() === null ) {
```
nickpelton, line 12: This is odd, empty should return true for NULL. https://www.php.net/manual/en/function.empty.php#114267 

Should we check for empty OR null to make sure other general empty conditions are covered?

lindseyolson: Yes, good call. I will add the empty() check back in.

lindseyolson: I guess the "OR" won't work. The issue we are having is that the string of zero - `"0"` is returning true for being empty. So if we have an empty OR null, it still won't pass.

nickpelton: Could we cast the value to an int prior the check if we always assume it's a number?

lindseyolson: Well it's technically a text field, so it can be anything (can have a percentage sign or I assume it could have a string value if they wanted to)

lindseyolson: I could explicitly just check that it's not equal to a string of 0, since that is the only issue we are having.

lindseyolson: `if ( empty( $c->get_content() ) || $c->get_content() !== '0' )`

nickpelton: Sounds good!

defunctl: > `if ( empty( $c->get_content() ) && $c->get_content() !== '0' )`

This will still fail because of the order that PHP processes statements (left to right). 

You'll need to reverse the check to check for the '0' string first.

defunctl: Actually, scratch that. This will work well.

nickpelton (approved):

PR #728, opened 2021-05-20
Fix husky hooks phpcs from checking every PHP file on commit

defunctl:
## What does this do/fix?

Stops husky hooks from running phpcs on every PHP file in the project and only looks in our core plugin/theme.

## QA

- Do some bad PHP in either `wp-content/plugins/core/ ` or `wp-content/themes/core/` and it will find it.
- Add a premium plugin to `wp-content/plugins` and it will no longer scan it.

## Tests

Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests because it's a package.json/tool update


nickpelton (approved):
🚀

PR #730, opened 2021-05-20
add dependabot dependencies management to SQ1.

nickpelton:
## What does this do/fix?

Adds the now native Github Dependabot config. This provides some nice benefits:

1. Automatic PRs for dependency updates
2. Schedule dependabot checks interval
3. PR tagging for dependency updates. 

Can read more here: https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/about-dependabot-version-updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,7 @@
 version: 2
 
 updates:
-  - package-ecosystem: "composer1"
+  - package-ecosystem: "composer"
     directory: "/"
     schedule:
       interval: "weekly"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,20 @@
+version: 2
+
+updates:
+  - package-ecosystem: "composer"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+    labels:
+      - "composer"
+      - "dependencies"
+
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+    labels:
+      - "npm"
+      - "dependencies"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,7 @@
 version: 2
 
 updates:
-  - package-ecosystem: "composer"
+  - package-ecosystem: "composer1"
     directory: "/"
     schedule:
       interval: "weekly"
--- a/composer.json
+++ b/composer.json
@@ -73,6 +73,7 @@
     }
   ],
   "require": {
+    "composer-plugin-api": "^1.1",
     "php": "^7.4",
     "advanced-custom-fields/advanced-custom-fields-pro": "*",
     "composer/installers": "1.4.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "1672fe34d5bbc4b30e08fbe3f4b20bff",
+    "content-hash": "34a5d06010775f309862e5477e1deae9",
     "packages": [
         {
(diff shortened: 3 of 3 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,20 @@
+version: 2
+
+updates:
+  - package-ecosystem: "composer"
```
defunctl, line 4:
```suggestion
  - package-ecosystem: "composer1"
```
I think this will force composer v1 which is what we need right now if it defaults to v2

defunctl (changes requested):

.github/dependabot.yml
nickpelton: Looking at their docs it was not very clear how to do this. Found this https://github.com/dependabot/dependabot-core/pull/2838#issuecomment-744436945 and it notes it should auto-detect based on a composer.lock composer-plugin-api setting, which look like we can now set as part of composer.json.

So instead, I _think_ we should add this to our composer.json.

```"require": {
        "composer-plugin-api": "^1.10"
    },
    "require-dev": {
        "composer/composer": "^1.10"
    },```

Thoughts?

Comment (defunctl) [reply .github/dependabot.yml:4]: I'm game with that if it works

**Comment by defunctl** [comment]
@nickpelton I pulled this down and ran it with composerv2 and it fails, wondering if we should just tell dependabot to use composer1 with the change in my comment.

```
Your lock file does not contain a compatible set of packages. Please run composer update.

  Problem 1
    - Root composer.json requires composer-plugin-api ^1.1, found composer-plugin-api[2.0.0] but it does not match the constraint.
  Problem 2
    - composer/installers is locked to version v1.4.0 and an update of this package was not requested.
    - composer/installers v1.4.0 requires composer-plugin-api ^1.0 -> found composer-plugin-api[2.0.0] but it does not match the constraint.
  Problem 3
    - ffraenz/private-composer-installer is locked to version v3.0.1 and an update of this package was not requested.
    - ffraenz/private-composer-installer v3.0.1 requires composer-plugin-api ^1.0.0 -> found composer-plugin-api[2.0.0] but it does not match the constraint.
  Problem 4
    - johnpbloch/wordpress-core-installer is locked to version 1.0.0 and an update of this package was not requested.
    - johnpbloch/wordpress-core-installer 1.0.0 requires composer-plugin-api ^1.0 -> found composer-plugin-api[2.0.0] but it does not match the constraint.
  Problem 5
    - dealerdirect/phpcodesniffer-composer-installer is locked to version v0.5.0 and an update of this package was not requested.
    - dealerdirect/phpcodesniffer-composer-installer v0.5.0 requires composer-plugin-api ^1.0 -> found composer-plugin-api[2.0.0] but it does not match the constraint.
  Problem 6
    - ffraenz/private-composer-installer v3.0.1 requires composer-plugin-api ^1.0.0 -> found composer-plugin-api[2.0.0] but it does not match the constraint.
    - gravityforms/gravityforms 2.4.23 requires ffraenz/private-composer-installer ^3.0 -> satisfiable by ffraenz/private-composer-installer[v3.0.1].
    - gravityforms/gravityforms is locked to version 2.4.23 and an update of this package was not requested.

You are using Composer 2, which some of your plugins seem to be incompatible with. Make sure you update your plugins or report a plugin-issue to ask them to support Composer 2.
```

**Comment by nickpelton** [comment]
I didn't see any docs that confirmed your change, but I'll swap it. Guess we'll never know till it runs.

**Comment by nickpelton** [comment]
Yeah, that fails: 
```
The property '#/updates/0/package-ecosystem' value "composer1" did not match one of the following values: npm, bundler, composer, maven, mix, cargo, gradle, nuget, gomod, docker, elm, gitsubmodule, github-actions, pip, terraform
```

defunctl:
> I didn't see any docs that confirmed your change, but I'll swap it. Guess we'll never know till it runs.

I read it in a PR but you are probably right, they themselves could be detecting the version by reading the `composer.json` themselves. But we should use the `1.0` of the API since our plugins still require that.

defunctl (approved):
:rocket:

PR #740, opened 2021-05-21
[SQONE-605] Minimal MCE toolbar (default for all custom blocks)

lindseyolson:
## What does this do/fix?
This PR adds a custom tiny MCE toolbar: 'minimal'

Minimal Toolbar includes:
- bold
- italic
- unordered list
- ordered list
- link

All custom blocks were updated to use the minimal toolbar as the default toolbar in the wysiwygs, along with removing media upload buttons and the "text" tab for the wysiwyg.


wp-content/plugins/core/src/Admin/Editor/Classic_Editor_Formats.php
```diff
@@ -110,4 +110,25 @@ public function visual_editor_styles_dropdown( $settings ): array {
...
+	 * Add minimal toolbar to use as default editor with custom blocks
+	 *
+	 * @filter acf/fields/wysiwyg/toolbars
+	 *
+	 * @param array $toolbars
+	 *
+	 * @return array
+	 */
+	public function add_minimal_toolbar( array $toolbars ): array {
+		$toolbars['Minimal'][1] = [
```
nickpelton, line 123: See this is Uppercase, while the call to use it is lowercase. Is it possible to keep this consistent?

Also, since we have a ClassicEditor class, could we set the toolbar options as constants we can use so we can avoid typos?

wp-content/plugins/core/src/Blocks/Types/Content_Loop/Content_Loop.php
```diff
@@ -100,25 +103,25 @@ public function add_fields() {
...
-					'multiple'        => 0,
-					'image_path'      => sprintf(
-						'%sassets/img/admin/blocks/%s/',
-						trailingslashit( get_template_directory_uri() ),
-						self::NAME
-					),
-					'image_extension' => 'svg',
-				] )
-			);
+			 ->add_field(
```
nickpelton, line 106: Weird green marks here, I checked the coded and it's not anything PHPStorm sees.. so ignore?

wp-content/plugins/core/src/Admin/Editor/Classic_Editor_Formats.php
lindseyolson: Yes and yes. By ClassicEditor class, you mean the current class, right? `Classic_Editor_Formats`?

nickpelton: Yup!  We call it `$toolbars['Minimal']` here and `'toolbar'      => 'minimal',` later. If it was a class constant, it would be simple to reference and would avoid typos. like
```
$toolbars[Classic_Editor_Formats::minimal]
```

I just wasn't sure the the Uppercase and lowercase difference was a requirement.

lindseyolson: I understand now, thanks for the clarification. I will update the const and add to the blocks now.

nickpelton (approved):
🚀

PR #743, opened 2021-05-24
update dependabot config to point at the correct branch

nickpelton:
## What does this do/fix?

Original Dependabot setup fails to determine the default branch. Added to configuration.

Docs: https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#target-branch


defunctl (approved):
:rocket:

PR #744, opened 2021-05-24
Feature/SQONE-281/Search-Template

tjdunklee:
## What does this do/fix?

This styles and adds additional functionality for the search template. Here are some details about added functionality:

- Adds a clear button to search form
- Adds new 3:2 aspect ratio for images (includes shim)
- Blocks search from searching on empty queries `?s=` (Thanks for that from Codet Dave!)
- Adds Card JS to search result items to make whole card clickable yet selectable
- Removes search form from header because we typically replace it with a direct link

## QA

Links to relevant issues
- [JIRA Ticket](https://moderntribe.atlassian.net/browse/SQONE-281)

Screenshots/video:
- Responsive Demo GIF: http://p.tri.be/x1mNza
- Desktop Results: http://p.tri.be/X5jBjF
- Mobile Results: http://p.tri.be/OKOvao
- Desktop No Results: http://p.tri.be/ZIRlQ6

## Tests

Does this have tests?

- [ ] Yes
- [ x] No, because I'm not sure it needs any. Maybe?


wp-content/plugins/core/src/Theme/Config/Image_Sizes.php
```diff
@@ -17,6 +17,11 @@ class Image_Sizes {
 	public const FOUR_THREE_LARGE       = 'four-three-large';
 	public const SOCIAL_SHARE_OPENGRAPH = 'social-share-opengraph';
 	public const SOCIAL_SHARE_TWITTER   = 'social-share-twitter';
+	public const THREE_TWO_XSMALL       = 'three-two-xsmall';
+	public const THREE_TWO_SMALL        = 'three-two-small';
+	public const THREE_TWO              = 'three-two';
+	public const THREE_TWO_LARGE        = 'three-two-large';
+	public const THREE_TWO_XLARGE       = 'three-two-xlarge';
```
dpellenwood, line 24: Hmmm... Let's chat this through with design. I don't know if I'm keen on adding more image sizes to SquareOne by default. Each set of sizes we define adds additional intermediate sizes that have to be generated when an image is added. This creates a ton of overhead. 

Matt CR or @kyleunzicker , would it be possible to use one or two set of images sizes (aspect ratos) throughout the defaults in SquareOne? Of course we can adjust ARs for each project. I'm just hoping to avoid creating ~20+ sizes on ever new project.  Thoughts?

wp-content/themes/core/components/loop_items/search/search.php
```diff
@@ -6,26 +6,30 @@
 $c = Search_Controller::factory();
 ?>
 
-<article class="item-loop item-loop--search item-loop--<?php echo esc_attr( get_post_type() ); ?>">
+<article class="search-result search-result--<?php echo esc_attr( get_post_type() ); ?>" data-js="use-target-link">
```
dpellenwood, line 9: To be totally honest, I'd prefer to see a card component used here instead of a separate template item. It creates a nice abstraction of loop items being cards that could be reused in any context (not "locked" into the results template).

Codet and Durham both have their results setup that way, if you need a reference. Let me know if you want to chat it through together.

wp-content/themes/core/components/routes/search/Search_Controller.php
```diff
@@ -69,14 +71,47 @@ public function render_breadcrumbs(): void {
...
+				__( 'Your search for &lsquo;%s&rsquo; returned 0 results', 'tribe' ),
+				$query_term
+			);
+		} else {
+			$text = sprintf(
+				_n( 'Showing %d result for &lsquo;%s&rsquo;', 'Showing %d results for &lsquo;%s&rsquo;', $total, 'tribe' ),
+				$total,
+				$query_term
+			);
+		}
```
dpellenwood, line 109:
```suggestion
		$text = sprintf(
			_n( 'Showing %d result for &lsquo;%s&rsquo;', 'Showing %d results for &lsquo;%s&rsquo;', $total, 'tribe' ),
			$total,
			$query_term
		);

		if ( 0 === $total ) {
			$text = sprintf(
				__( 'Your search for &lsquo;%s&rsquo; returned 0 results', 'tribe' ),
				$query_term
			);
		}
```

wp-content/themes/core/components/routes/search/Search_Controller.php
```diff
@@ -69,14 +71,47 @@ public function render_breadcrumbs(): void {
...
+			Search_Form_Controller::VALUE   => get_search_query(),
+		];
+
+		return $args;
+	}
+
+	/**
+	 * @return string
+	 */
+	public function get_results_text_args() {
```
dpellenwood, line 89: Don't forget the return type: `public function get_results_text_args(): array {`,

Additionally it appears that the docblock return type is wrong.

wp-content/themes/core/components/routes/search/css/search.pcss
```diff
@@ -0,0 +1,114 @@
...
+
+/* -----------------------------------------------------------------------------
+ * Search Results: Summary
+ * ----------------------------------------------------------------------------- */
+
+.search-results__summary {
+	margin-bottom: var(--spacer-20);
+	line-height: 1.75;
+	font-weight: 700;
+	letter-spacing: 0.1px;
```
dpellenwood, line 38: Any chance we can use an existing typography mixin for this?

If this really is a unique style and it needs to be unique, let's add a mixin and helper class for it in the typography pcss partials. We're trying to avoid any one-off typeography throughout the components. Additionally, please run it by design to make sure it gets added to the Figma Style Guide for typography styles.

wp-content/themes/core/components/routes/search/css/search.pcss
```diff
@@ -0,0 +1,114 @@
...
+
+.search-result__image {
+	display: none;
+
+	@media (--viewport-medium) {
+		display: block;
+		width: 168px;
+		flex-shrink: 0;
+	}
+}
```
dpellenwood, line 114: These can all be part of the a specific card style/variation once the loop-item template is consolidated into a card style.

dpellenwood (changes requested):
Really great work, Tim. Some changes and suggestions. Hit me up if you have any questions.

wp-content/plugins/core/src/Query/Query_Subscriber.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Tribe\Project\Query;
+
+use Tribe\Libs\Container\Abstract_Subscriber;
+
+class Query_Subscriber extends Abstract_Subscriber {
+	public function register(): void {
+		add_action( 'posts_request', function ( $request, $wp_query ) {
```
defunctl, line 9: This is technically a filter and should use `add_filter` here.

wp-content/plugins/core/src/Query/Search.php
```diff
@@ -0,0 +1,33 @@
...
+		//sending back a 404 and still provides the page title, etc.
+		if ( $query->is_search() && ! is_admin() && empty( get_query_var( 's' ) ) ) {
+
+			/* prevent SELECT FOUND_ROWS() query */
+			$query->query_vars['no_found_rows'] = true;
+
+			/* prevent post term and meta cache update queries */
+			$query->query_vars['cache_results'] = false;
+
+			return false;
```
defunctl, line 27:
```suggestion
			return '';
```
This filter returns a string, so we should do the same and fix the return type.

wp-content/plugins/core/src/Query/Search.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+namespace Tribe\Project\Query;
+
+class Search {
+	/**
```
defunctl, line 6: Once you merge in `main` and run `yarn install; lefthook install -a` phpcs should catch the formatting issues in this file.

wp-content/plugins/core/src/Query/Search.php
```diff
@@ -0,0 +1,33 @@
...
+	 * @param           $request
+	 * @param \WP_Query $query
+	 *
+	 * @return bool
+	 * @hook https://developer.wordpress.org/reference/hooks/posts_request/
+	 */
+	public function dont_do_search( $request, \WP_Query $query ) {
+		//This basically prevents the query for search but still provides the benefit of not
+		//sending back a 404 and still provides the page title, etc.
+		if ( $query->is_search() && ! is_admin() && empty( get_query_var( 's' ) ) ) {
```
defunctl, line 19:
```suggestion
		if ( $query->is_search() && ! is_admin() && empty( $query->get( 's' ) ) ) {
```

defunctl (changes requested):

wp-content/plugins/core/src/Query/Search.php
```diff
@@ -0,0 +1,33 @@
+<?php
```
ryanurban, line 1: Nothing wrong with handling this this way. Another way we've handled this in the past is using JS to prevent form submit if the input is empty, but I think that's also been on a more project by project basis.

wp-content/themes/core/components/loop_items/search/search.php
ryanurban: Agreed!

ryanurban (commented):
Nice work, Tim 🤘

tjdunklee:
Ready for re-review! I updated the following:

- Fixed PHP errors/warning
- Updated search result items to be a new card styles
- Consolidated typographic styles (less custom styles)

wp-content/themes/core/components/card/css/card.pcss
```diff
@@ -100,6 +118,17 @@
...
 		}
 	}
+
+	/* CASE: Style: Search */
+	.c-card--style-search & {
+		display: none;
+
+		@media (--viewport-medium) {
+			display: block;
+			width: 168px;
```
dpellenwood, line 128: I think you can use: `--grid-2-col` for this value (so it can automatically adjust if the grid size changes).

wp-content/themes/core/components/card/css/card.pcss
```diff
@@ -147,6 +189,18 @@
 			text-decoration: underline;
 		}
 	}
+
+	/* CASE: Style: Search */
+	.c-card--style-search & {
+		@mixin t-input;
+		margin-bottom: var(--spacer-10);
+		font-weight: 700;
```
dpellenwood, line 197: While this does match the designs, let's adjust it to `@mixin t-display-xx-small;` to keep it consistent with other "display" or heading styles.

wp-content/themes/core/components/card/css/card.pcss
```diff
@@ -179,4 +243,27 @@
...
+
+/* -----------------------------------------------------------------------------
+ * Card Call To Action Link
+ * ----------------------------------------------------------------------------- */
+
+.c-card__cta-link {
+
+	/* CASE: Style: Search */
+	.c-card--style-search & {
+		word-break: break-all; /* Break long URLs */
```
dpellenwood, line 261: Noice!

wp-content/themes/core/components/card/css/card.pcss
```diff
@@ -179,4 +243,27 @@
...
+
+		&:focus,
+		&:hover,
+		.c-card--style-search:hover& {
```
dpellenwood, line 265: I _think_ there's a space missing before the ampersand?

dpellenwood (approved):
Couple minor things, then good to merge. Thanks, Tim!

wp-content/themes/core/components/card/css/card.pcss
tjdunklee: Yea, this selector is kind of weird, but I had to do that to access the hover state of the parent `.c-card---style-search`. It builds a funky selector, but it is nice to keep the attribute change on one line. Any thoughts of how to clean that up @dpellenwood?

dpellenwood: Ah, gotcha. You're g2g with this as-is. Maybe just a comment for future-us?

wp-content/plugins/core/src/Query/Search.php
```diff
@@ -0,0 +1,34 @@
...
+
+namespace Tribe\Project\Query;
+
+class Search {
+
+	/**
+	 * Prevents the query for the WP Core search from running when there is no search term `?s=`,
+	 * but still returns the search route (page title, template, etc).
+	 *
+	 * @param           $request
```
defunctl, line 11: Add `string` here

wp-content/plugins/core/src/Query/Search.php
```diff
@@ -0,0 +1,34 @@
...
+	 * @param \WP_Query $query
+	 *
+	 * @return bool
+	 *
+	 * @hook https://developer.wordpress.org/reference/hooks/posts_request/
+	 */
+	public function dont_do_search( $request, \WP_Query $query ) {
```
defunctl, line 18: Can we rename this to something like `force_load_search_template()`?

defunctl (changes requested):
Nice, a couple more typing fixes (which the current rules aren't detecting but will be later, so might as well not have to fix it later)

wp-content/plugins/core/src/Query/Search.php
tjdunklee: Good idea!

wp-content/themes/core/components/routes/search/Search_Controller.php
```diff
@@ -68,15 +73,121 @@ public function render_breadcrumbs(): void {
...
 
 		return [
-			new Breadcrumb( $url, __( 'News', 'tribe' ) ),
+			Text_Controller::TAG     => 'p',
+			Text_Controller::CLASSES => [ 't-display-xx-small', 'search-results__summary' ],
+			Text_Controller::CONTENT => esc_html( $text ),
+		];
+	}
+
+	/**
```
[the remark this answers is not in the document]
tjdunklee: I'm confused what you are referring to here. Also, after I added your code suggestion of a return type, I'm getting a fatal error. Sorry, I don't really understand any of this type stuff in PHP. http://p.tri.be/gTWutu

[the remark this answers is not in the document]
tjdunklee: I took a stab at fixing this issue. Let me know what you think @defunctl!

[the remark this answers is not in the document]
defunctl: @tjdunklee looks like this broke the search acceptance tests. I'll take a look at that.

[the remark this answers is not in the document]
tjdunklee: Yea, this whole PR from the beginning broke that test. I know exactly what is happening, I just have no idea how to write a test. It is breaking because we removed the search form from the header. The search form is only on the search template now, so we should just be testing if the search works there.

PR #755, opened 2021-05-27
[SQONE-603] Limit CI tests to only PR's against the main branch

nickpelton:
## What does this do/fix?

Makes CI run on the main branch for Push and Pull_Request. This should fix our  badge and also reduce unnecessary test runs.


defunctl (approved):
:rocket: :moon:

PR #764, opened 2021-06-01
Proper wp_nav_menu() $args type

defunctl:
## What does this do/fix?

Getting ahead of another casting issue we ran into on another project.

WordPress claims the $args are an array, but they cast it as an object in a function.

## QA

- Menus on the frontend will no longer throw an error once types are added. 

## Tests

Does this have tests?

- [ ] Yes
- [x] No

### DIFF
--- a/wp-content/plugins/core/src/Nav_Menus/Nav_Attribute_Filters.php
+++ b/wp-content/plugins/core/src/Nav_Menus/Nav_Attribute_Filters.php
@@ -2,6 +2,8 @@
 
 namespace Tribe\Project\Nav_Menus;
 
+use stdClass;
+
 /**
  * Class Nav_Attribute_Filters
  *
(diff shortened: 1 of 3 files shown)


nickpelton (approved):
Interesting it was failing since $arg was casted in the method. No objections to moving it. Approved.

defunctl:
> Interesting it was failing since $arg was casted in the method. No objections to moving it. Approved.

It's not casted in the method, the expectation is defined in the method.

Actually, I think they fixed this and we're now just passing the [wrong type](https://github.com/WordPress/WordPress/blob/5.7-branch/wp-includes/class-walker-nav-menu.php#L199)

defunctl:
@nickpelton updated this to properly match WordPress now.

nickpelton:
I meant how it was casting it inside the method just like you did on the subscriber. `$args = (array) $args;`. But yeah, they look to have fixed it.

nickpelton (approved):

PR #769, opened 2021-06-03
[SQONE-619] Update Interstitial Block Fields

lindseyolson:
## What does this do/fix?
Updates/Adds the Field Improvements for the Interstitial Block Fields.


wp-content/themes/core/components/blocks/interstitial/Interstitial_Block_Controller.php
```diff
@@ -142,6 +147,19 @@ public function get_content_args(): array {
...
 	}
 
+	/**
+	 * @return \Tribe\Project\Templates\Components\Deferred_Component
+	 */
+	private function get_leadin(): Deferred_Component {
+		return defer_template_part( 'components/text/text', null, [
+			Text_Controller::CLASSES => [
+				'c-block__leadin',
+				'b-tabs__leadin',
```
ryanurban, line 157: I think you can nix this. This looks like a class specific to the tabs block

ryanurban (approved):
🤘

wp-content/themes/core/components/blocks/interstitial/Interstitial_Block_Controller.php
lindseyolson: Good catch, thank you!

PR #773, opened 2021-06-04
Can't make this a required check with paths

defunctl:
## What does this do/fix?

If you make the eslint action part of `Require status checks to pass before merging` for a branch and no files are changed, it never runs and therefore never completes and sits waiting forever. The only solution I could come up with is to make it run, as it already filters which files it runs on.

## Tests

Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests because it's a GitHub action.


nickpelton (approved):
Approved. Should we also make this run on push to main so we check post merge?

defunctl:
> Approved. Should we also make this run on push to main so we check post merge?

Great idea, I'll update it.

nickpelton (approved):
🚀

dpellenwood (approved):
Nice catch, Justin.

nickpelton:
Hey @defunctl ready to merge.

PR #775, opened 2021-06-04
Feature/sqone 281/be search template

defunctl:
## What does this do/fix?

- Removes broken call to get_breadcrumbs() method.
- Makes Search Controller extend the Index controller.
- Updates acceptance tests to test this feature instead of failing now that the search bar is removed from all pages.

## Tests

Does this have tests?

- [x] Yes


tjdunklee (approved):
Looks good to me @defunctl! Thanks for jumping in and fixing these tests. I didn't actually run the test myself because I haven't used Docker for a while, but I guess we'll see if it passes in the PR.

nickpelton (approved):
🚀

PR #777, opened 2021-06-05
Patch utils not being watched by webpack, add a11y loop module

faction23:
## What does this do/fix?

This pr patches an issue with watch for webpack not firing for utils, and also adds a new common directory for js in anticipation of js modules that get shared between admin and theme that aren't apps, with FSE and all coming down the road. 

This also adds an accessibility module called focusLoop, which allows you to apply the event trapping pattern used in our a11y dialog on any element you like. Can be used like

```javascript
handleKeyEvents = ( e ) => focusLoop( e, trigger, container );

container.addEventListener( 'keydown', this.handleKeyEvents );
```


webpack/plugins/watch.js
```diff
@@ -1,17 +1,32 @@
+const path = require( 'path' );
 const MiniCssExtractPlugin = require( 'mini-css-extract-plugin' );
+const WatchExternalFilesPlugin = require( 'webpack-watch-files-plugin' ).default;
```
dpellenwood, line 3: Curious about the `.default` usage here?

dpellenwood (approved):
Looks great, Sam. Just one question. Thank you!

webpack/plugins/watch.js
faction23: It's the way this plugin exports, def a quirk. If they update at some point we may have to change this!

PR #781, opened 2021-06-07
[SQONE-614] Update Media Text Block Fields

lindseyolson:
## What does this do/fix?
Updates/Adds the Field Improvements for the Media Text Block Fields.


ryanurban (approved):
🤘

PR #783, opened 2021-06-07
Updating the eslint file to include JS in the component directory

ScottGarrison:
## What does this do/fix?

- Adds in the component JS files to our eslint command
- Ran it for testing and it includes the dialog JS file updates.


nickpelton (approved):
Looks good. I'm surprised the theme check with `**/*` doesn't pick up the component folders. I thought `**/*` was essentially recursive folder checks. 🤷

ScottGarrison:
Only within the root structure. None of our checks currently had the root outside of the JS folder

nickpelton:
Do we have the same issue with the stylelint? It also is scoped to only look directly into the `wp-content/themes/core/assets/css/src/theme/` folder. So it wouldn't be checking the components properly either, correct? Can we fix that as part of this PR as well?

dpellenwood:
> Do we have the same issue with the stylelint? ... Can we fix that as part of this PR as well?
Agreed. Let's add the components path to stylelint as well.

I'd suggest we should add the integrations folder path to both eslint and stylelint as well.

dpellenwood (changes requested):
Changes look great. Please add integrations to eslint and component & integrations paths to the theme stylelint task.

ScottGarrison:
@nickpelton @dpellenwood I made the requested changes but the Codeception test is failing. Seems like it's running tests or lint on files that don't exist, and I cant even find the variables it's calling out in our file structure. Must be something running i'm not familiar with

ScottGarrison:
@nickpelton @dpellenwood This is good for a recheck with the added lint for the component css and integrations

dpellenwood (approved):
Awesome! Thank you Scott!

PR #787, opened 2021-06-08
Feature/phpcs workflow updates

defunctl:
## What does this do/fix?

- Currently if you have the `sniff` job from the PHPCS workflow set under "Require status checks to pass before merging" and no PHP files changed in our specified paths, the PR will not pass those checks and the PR can only be merged by administrators. 
- Uses [dorny/paths-filter ](https://github.com/dorny/paths-filter) action to detect file changes instead so the workflow will still pass if no PHP file changes are present in the PR.

## QA

**Test with no PHP file changes**
- Ensure the `sniff` job is added to "Require status checks to pass before merging"  under Settings > Branches > main > Protect matching branches.
- Change only a JS file and open a PR.
- The PHPCS workflow will now properly pass.

**Test PHP file changes in specified paths**
- Ensure the `sniff` job is added to "Require status checks to pass before merging"  under Settings > Branches > main > Protect matching branches
- Make a breaking PHPCS change to a PHP file under `wp-content/plugins/core/`, and/or `wp-content/themes/core/` and/or `wp-content/mu-plugins` and the job should properly fail.

## Tests

Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests because it's a GitHub workflow change


nickpelton (changes requested):
Passed the QA tests. 

One proposed change, since this is static analysis of changed files only, we don't need a composer build do we? Can trim off 3-4m of CI time by removing that.

defunctl:
> Passed the QA tests.
> 
> One proposed change, since this is static analysis of changed files only, we don't need a composer build do we? Can trim off 3-4m of CI time by removing that.

We can't run phpcs without running composer install since that's what puts the phpcs binary in the `vendor/bin` folder: https://github.com/moderntribe/square-one/pull/787/files#diff-f621c7ec73ae73c758a64d5f030f2062fc73f50676d953035650636c052aac3fR86

nickpelton (approved):
Ah. OK lets ship as is. I'll think about how we could do a partial install in the future to avoid a full build.

PR #789, opened 2021-06-08
Bug/links block rendering

nickpelton:
## What does this do/fix?

1. The Links block renders nothing without list data. There is a content block above so this is confusing to the user as they add content and see nothing. All content added will now render.
2. The Link block uses nested repeaters that don't pass our null check in our Model->get method.  Why? ACF returns an array of data from repeaters, and the keys can have a null value for nested fields and in this case the links. This fixes the issue at the link controller level by checking for a null value and casting it back to a string.
3. Fixed an illegal offset warning error
4. Adjusted bullet character alignment for chrome issue.

## QA

Testing Instructions: 
1. Goto test link below.
2. VERIFY all content renders.
3. VERIFY when adding a new link to the repeater with no link added. Post Saved. No error.

Test Link:
- https://sq1-pr789.moderntribe.qa/wp-admin/post.php?post=7&action=edit

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/SQONE-629)

### DIFF
--- a/wp-content/themes/core/components/blocks/links/css/links.pcss
+++ b/wp-content/themes/core/components/blocks/links/css/links.pcss
@@ -5,12 +5,10 @@
  * ----------------------------------------------------------------------------- */
 
 .b-links {
-
 	/* CASE: Layout Stacked is slightly different than content block inline layout */
 	.c-content-block--layout-inline {
 
 		.c-block__title {
-
 			@media (--viewport-medium) {
 				float: none;
 				width: 100%;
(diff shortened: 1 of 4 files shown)


wp-content/themes/core/components/blocks/links/css/links.pcss
```diff
@@ -171,7 +169,7 @@
 
 		position: absolute;
 		left: 0;
-		top: 4px;
+		top: 0;
```
dpellenwood, line 172: Curious about this change?

dpellenwood (approved):
Small question, but looks great. Thanks Nick!

wp-content/themes/core/components/blocks/links/css/links.pcss
nickpelton: The psudo element cheveron bullets were misaligned. This fixes that.

PR #792, opened 2021-06-10
Feature/sqone 554 allow block filter

mmcachran:
## What does this do/fix?

Adds a filter for allowed Gutenberg blocks.

Example:
```
add_filter(
	'tribe_allowed_blocks',
	function ( $allowed_blocks ) {
		$allowed_blocks    = is_array( $allowed_blocks ) ? $allowed_blocks : [];
		$allowed_blocks[] = 'core/block';
		return $allowed_blocks;
	},
	10,
	1
);
```


nickpelton (approved):
Filter code looks fine. Approved.

Question for thought, should we instead be filtering the list in the definer where it's created rather than downstream in this method? I could see some confusion where you add something to the definer list and not know why it's not appearing downstream. But it also think it goes both ways. Maybe that doesn't matter, as we should probably be making that list shorter and filtering in each block upon registration.

nickpelton (approved):

PR #799, opened 2021-06-17
Feature/sqone 193 breadcrumbs

mmcachran:
## What does this do/fix?

Adds default breadcrumb functionality and docs.

## QA

Links to relevant issues
- [JIRA Ticket](https://moderntribe.atlassian.net/browse/SQONE-193)


wp-content/themes/core/components/breadcrumbs/breadcrumbs.php
```diff
@@ -14,15 +14,25 @@
 	<?php echo $c->get_attrs(); ?>
 >
 
-	<ul <?php echo $c->get_main_classes(); ?> <?php echo $c->get_main_attrs(); ?>>
+	<ul itemscope itemtype="http://schema.org/BreadcrumbList" <?php echo $c->get_main_classes(); ?> <?php echo $c->get_main_attrs(); ?>>
```
nickpelton, line 17: Nice bringing in a schema here!

wp-content/plugins/core/src/Templates/Components/Abstract_Controller.php
```diff
@@ -59,4 +71,96 @@ public static function factory( array $args = [] ) {
 		return tribe_project()->container()->make( static::class, [ 'args' => $args ] );
 	}
 
+	public function render_breadcrumbs(): void {
```
nickpelton, line 74: This is more of an opinion, but I think we shouldn't load templates within the controller and just add this call to the template.

wp-content/plugins/core/src/Templates/Components/Abstract_Controller.php
```diff
@@ -59,4 +71,96 @@ public static function factory( array $args = [] ) {
...
+			]
+		);
+	}
+
+	/**
+	 * Returns breadcrumbs for the post.
+	 *
+	 * @return array \Tribe\Project\Templates\Models\Breadcrumb[] Breadcrumbs for the post.
+	 */
+	protected function get_breadcrumbs(): array {
```
nickpelton, line 89: All this breadcrumb data logic, should this be a separate class, or a Trait? Now all controllers have breadcrumb functionality, which we don't need for many components (Like the button component for example). We really only need them are part of top level  (Page) components.

nickpelton (commented):
Comments for discussion.

PR #824, opened 2021-07-14
Add require-dev symfony/laravel dd package

nickpelton:
## What does this do/fix?

Adds Symfony/Laravel `dd()` command for quick debugging needs via require-dev. Often when I just need a quick dump I find myself wanting the convenience of the `dd()` command from Laravel. I propose we add it to squareone dev dependencies.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -121,6 +121,7 @@
     "codeception/module-webdriver": "^1.0",
     "codeception/util-universalframework": "^1.0",
     "filp/whoops": "^2.2@dev",
+    "larapack/dd": "^1.1",
     "lucatume/function-mocker": "~1.0",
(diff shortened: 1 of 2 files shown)


dpellenwood (approved):
Tbh, I tend to stick to xDebug breakpoints, but I have no feelings one way or the other.

nickpelton:
Me too. But sometimes this is a nice alternative if I don't wanna take the time to enable xdebug (I often have it off for performance on mac).

ChrisMKindred (approved):
Anything that helps to debug I am all for. Is it overkill to include a package for the 5 lines of code?

defunctl (approved):

PR #836, opened 2021-07-22
Feature/wp5.8

nickpelton:
## What does this do/fix?

* Update to WP5.8
* Update Allowed Blocks filter for latest deprecation
* Disable block widgets till we choose to support them


dpellenwood (approved):
:bus: :dash:

PR #837, opened 2021-07-22
Block Deny List

ChrisMKindred:
## What does this do/fix?

Removes the Allow Blocks and replaces it with the ability to Deny blocks you don't want available to the user. 

## QA

Dokku failed to deploy on the PR, so I created a quick video:
- [Deny List In Action](http://p.tri.be/d4TSpN)

## Tests

Does this have tests?

- [ ] Yes


wp-content/plugins/core/src/Blocks/Deny_Blocks.php
```diff
@@ -0,0 +1,34 @@
+<?php declare(strict_types=1);
+
+namespace Tribe\Project\Blocks;
+
+/**
+ * Class Deny_Blocks
+ *
+ * Provides an array of block names to remove from the editor.
+ */
+class Deny_Blocks {
```
defunctl, line 10: Should name this as an object e.g. noun/noun phrase. Ask yourself, "Can you hand me the <class name>?" and does it make grammatical sense? 

Can you hand me the Deny Blocks?

How about `Block_Denier`?

wp-content/plugins/core/src/Blocks/Deny_Blocks.php
```diff
@@ -0,0 +1,34 @@
...
+
+	/**
+	 * @var array A list of blocks types to disable
+	 */
+	private $denylist;
```
defunctl, line 15:
```suggestion
	private array $denylist;
```

wp-content/plugins/core/src/Blocks/Blocks_Subscriber.php
```diff
@@ -19,12 +19,12 @@ public function register(): void {
...
-		add_filter( 'allowed_block_types', function ( $types, $post ) {
-			return apply_filters(
-				'tribe_allowed_blocks',
-				(array) $this->container->get( Allowed_Blocks::class )->register_allowed_blocks( $types, $post ),
-			);
-		}, 10, 2 );
+		/**
+		 * Adds the deny list to the JS_Config class.
+		 */
+		add_filter( 'tribe/project/blocks/denylist', function ( $types ) {
```
defunctl, line 25:
```suggestion
		add_filter( 'tribe/project/blocks/denylist', function ( array $types ): array {
```

defunctl (changes requested):
I'll reach out and we can discuss a few simple tests we could write for this.

Thanks, Chris!

wp-content/plugins/core/src/Blocks/Deny_Blocks.php
ChrisMKindred: `Denied_Blocks` maybe? That keeps with the previous convention of `Allowed_Blocks` and could help guide other developers that have been using the `Allowed_Blocks` functionality that is being removed with this PR.

defunctl: Allowed Blocks was also an improper name. There shouldn't be any plural classes. 

An opinionated article on the subject:
https://medium.com/swlh/an-opinionated-guide-to-naming-your-code-aimed-at-new-developers-6f769d60e098

How about `Block_Deny_List::add( $types )`?

That should be clear to any developer new and old what's happening at a glance

ChrisMKindred: Yes, `Block_Deny_List` is perfect.  Thanks. updating it now.

defunctl: Sorry was in a rush for an appointment before, but I also wanted to explain why this is important and why it makes sense.

While PHP-DI just does this behind the scenes, if you run `$deny_list = new Block_Deny_List( [] );` what is actually happening? 

You're creating an object (an instance of a class). Naming as a singular noun makes a lot more sense because you end up with a single instance of a class, say using `Animal()` as an example:

````php
$tiger = new Animal( 'tiger );
$bear = new Animal( 'bear' );
$chicken = new Animal( 'chicken' );
```
Would be strange to have a single instance of a plural object:

````php
// If it's plural, is this...a bunch of animals or what? It's just more confusing since you're creating a single instance of an Animal for each animal.
$tigers = new Animals( 'tiger' );
$bears = new Animals( 'bear' );
$chickens = new Animals( 'chicken' );
```

nickpelton (approved):
🚀

dpellenwood (approved):
This looks great! Thank you.

defunctl:
Chris is going to take a stab at writing some tests for this. This would be a good thing to have tests for in the event WordPress changes something in Gutenberg/Core and we update to the latest version it should catch if this is all of the sudden broken.

PR #849, opened 2021-08-09
Fix duplicate keys/broken quote example

defunctl:
## What does this do/fix?

- Fixes duplicate keys in the Quote block example so it renders the preview properly.

## QA

Block preview now shows values properly

image

## Tests

Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests because it's a block update.


nickpelton (approved):
:shipit:

PR #853, opened 2021-08-18
Bugfix for Comment Markup

ChrisMKindred:
## What does this do/fix?

When running the callback function in `Comments_Section_Controller::get_comments` we were returning the closing markup element.  The callback is being called from the `start_el` of the walker which means that we shouldn't be closing the element, the end_el will handle that. 

This PR also allows the comment controller to use the same comparison in order to build the element (li or div)to use based on the style in `wp_list_comments` the same way that the end_el method does for better compatibility.

## QA

Links to relevant issues
- [Link to Issue](https://example.com)

- [Link to Demo](https://example.com)

Screenshots/video:

## Tests

Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests.


ChrisMKindred:
Wrong base... will re-PR in a minute.

PR #854, opened 2021-08-18
fixes markup closing issue in comments

ChrisMKindred:
## What does this do/fix?
Fixes an invalid markup error in comments markup.

When running the callback function in `Comments_Section_Controller::get_comments` we were returning the closing markup element. The callback is being called from the `start_el` of the walker which means that we shouldn't be closing the element, the `end_el` will handle that.

This PR also allows the comment controller to use the same comparison in order to output the element (li or div) based on the style in `wp_list_comments` the same way that the `end_el` method does for better compatibility.

## QA

Screenshots/video:
- [invalid HTML example Current Output](http://p.tri.be/AsbKC1)
- [Fixed HTML after PR applied](https://sq1-pr854.moderntribe.qa/hello-world/)

## Tests

Does this have tests?

- [ ] Yes

### DIFF
--- a/wp-content/themes/core/components/comments/comment/Comment_Controller.php
+++ b/wp-content/themes/core/components/comments/comment/Comment_Controller.php
@@ -124,7 +124,7 @@ class Comment_Controller extends Abstract_Controller {
 		return date( $format, get_comment_time( 'U' ) );
 	}
 
-	public function get_starting_markup_element(): string {
+	public function get_tag(): string {
 		$element = 'li';
 		if ( 'div' === $this->style ) {
 			$element = 'div';
--- a/wp-content/themes/core/components/comments/comment/Comment_Controller.php
+++ b/wp-content/themes/core/components/comments/comment/Comment_Controller.php
@@ -14,17 +14,19 @@ class Comment_Controller extends Abstract_Controller {
 
 	use Comment_Edit_Link;
 
-	public const COMMENT_ID = 'comment_id';
-	public const CLASSES    = 'classes';
 	public const ATTRS      = 'attrs';
+	public const CLASSES    = 'classes';
+	public const COMMENT_ID = 'comment_id';
 	public const DEPTH      = 'depth';
 	public const MAX_DEPTH  = 'max_depth';
+	public const STYLE      = 'style';
 
-	private int $comment_id;
-	private array $classes;
 	private array $attrs;
+	private array $classes;
+	private int $comment_id;
 	private int $depth;
 	private int $max_depth;
+	private string $style;
 
 	/**
 	 * Comment constructor.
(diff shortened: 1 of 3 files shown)


wp-content/themes/core/components/comments/comment/Comment_Controller.php
```diff
@@ -120,4 +124,13 @@ public function get_time( $format = 'c' ): string {
 		return date( $format, get_comment_time( 'U' ) );
 	}
 
+	public function get_starting_markup_element(): string {
+		$element = 'li';
+		if ( 'div' === $this->style ) {
+			$element = 'div';
+		}
+
+		return $element;
```
RobertoGongora, line 133: This block of code could potentially be reduced to a single line, like this: 

```suggestion
		return ( 'div' === $this->style ) ? 'div' : 'li';
```

RobertoGongora (approved):
LGTM! Just added one suggestion

wp-content/themes/core/components/comments/comment/Comment_Controller.php
```diff
@@ -14,17 +14,19 @@ class Comment_Controller extends Abstract_Controller {
...
 	use Comment_Edit_Link;
 
-	public const COMMENT_ID = 'comment_id';
-	public const CLASSES    = 'classes';
 	public const ATTRS      = 'attrs';
+	public const CLASSES    = 'classes';
+	public const COMMENT_ID = 'comment_id';
 	public const DEPTH      = 'depth';
 	public const MAX_DEPTH  = 'max_depth';
+	public const STYLE      = 'style';
```
dpellenwood, line 22: In other components, we've opted to use the `TAG` const and `$tag` property to denote the HTML tag to use for a component. (Reference the Text block as a example.) I'd prefer to keep that syntax in the component as well for consistency.

wp-content/themes/core/components/comments/comment/Comment_Controller.php
dpellenwood, line 133: Rather than trying to "know" what wrapper tag we're going to use. I wonder if it would make more sense to add a attribute and property for a `COMMENT_TAG` or similar?

dpellenwood (commented):
Looks good! A suggestion and a question. :)

wp-content/themes/core/components/comments/comment/Comment_Controller.php
ChrisMKindred: This is actually the parent tag that is defined as the style attribute when calling the walker. I didn't want to add confusion since it isn't a 1:1 with the child tag.  The style attribute is either ul, ol, or div.

dpellenwood: Still, it's not a "style" it's an it's an html tag. I'd prefer to keep the component consistent with other components. If we're rendering a html tag, then let's use `TAG` and 

```
public function get_tag(): string {
	return tag_escape( $this->tag );
}
```

Instead of trying to deduce the tag based on a parent tag/markup.

wp-content/themes/core/components/comments/comment/Comment_Controller.php
dpellenwood: We wouldn't need this functionality at all if we just allow the correct tag to be passed into the component (defaulting to an `li` or a `div`).

ChrisMKindred: I think we should keep the conventions that the [`end_el` uses in the walker](https://github.com/WordPress/WordPress/blob/27522c6f677f022f0e5155db999e0e5fcbd794ac/wp-includes/class-walker-comment.php#L220). If we pass something different, the `end_el` is still going to check if the style arg is a div and write out the closing element depending on it.

dpellenwood: Eeew. Core Comments code. :grinning: 

Yep, You're right. Let's keep it consistent with core and stick to the `style` const & property.

One bit of compromise, can we call the method, `get_tag()` instead of `get_starting_markup_element()` so that at least _our_ components' method naming stays consistent?

nickpelton (approved):

PR #858, opened 2021-09-01
Fix/subheader2

nickpelton:
## What does this do/fix?

Fixes some inconsistencies in the background image handling across CPTs and the various data they provide.

### DIFF
--- a/wp-content/themes/core/components/header/subheader/Subheader_Controller.php
+++ b/wp-content/themes/core/components/header/subheader/Subheader_Controller.php
@@ -61,9 +61,10 @@ class Subheader_Controller extends Abstract_Controller {
 
 	protected function required(): array {
 		return [
-			self::MEDIA_CLASSES   => [ 'c-subheader__media' ],
-			self::CONTENT_CLASSES => [ 'c-subheader__content' ],
-			self::CLASSES         => [ 'c-subheader'],
+			self::CONTAINER_CLASSES => [ 'l-container' ],
+			self::MEDIA_CLASSES     => [ 'c-subheader__media' ],
+			self::CONTENT_CLASSES   => [ 'c-subheader__content' ],
+			self::CLASSES           => [ 'c-subheader'],
 		];
 	}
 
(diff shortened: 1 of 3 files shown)


wp-content/themes/core/components/header/subheader/Subheader_Controller.php
```diff
@@ -61,15 +61,16 @@ protected function defaults(): array {
 
 	protected function required(): array {
 		return [
-			self::MEDIA_CLASSES   => [ 'c-subheader__media' ],
-			self::CONTENT_CLASSES => [ 'c-subheader__content' ],
-			self::CLASSES         => [ 'c-subheader'],
+			self::CONTAINER_CLASSES => [ 'l-container' ],
+			self::MEDIA_CLASSES     => [ 'c-subheader__media' ],
+			self::CONTENT_CLASSES   => [ 'c-subheader__content' ],
+			self::CLASSES           => [ 'c-subheader', 'c-subheader--has-background'],
```
dpellenwood, line 67: Not sure we want to force `c-subheader--has-background`, do we? This seems like one we'd want to add in the `get_classes()` method.

wp-content/themes/core/components/header/subheader/css/subheader.pcss
```diff
@@ -25,13 +27,16 @@
...
 
 	@media (--viewport-medium) {
 		align-items: flex-end;
 	}
 }
 
+.c-subheader--has-image {
+	color: var(--color-white);
+}
+
```
dpellenwood, line 39: Instead of setting text color in this case, please add the css class `t-theme--light` to main subheader classes, or even just the content wrapper's classes.

dpellenwood (changes requested):
Couple small suggestions to better align with the system, but looks good overall.

wp-content/themes/core/components/header/subheader/Subheader_Controller.php
nickpelton: Ah good call. Fixed.

wp-content/themes/core/components/header/subheader/css/subheader.pcss
nickpelton: A neat. TIL.

nickpelton:
@dpellenwood Changes made. Please take another look.

dpellenwood (approved):
:duck: :dash:

PR #862, opened 2021-09-01
SQONE-521 Add a generic navigation menu component.

dpellenwood:
## What does this do/fix?
Adds a generic navigation menu component to SquareOne. This component will be used to disassociate the nav menu implementation from the context it appears in (masthead/footer/legal, etc). In future PRs I'll apply this component to SquareOne's menu locations.

[SQONE-521]
[SQONE-519]

[SQONE-521]: https://moderntribe.atlassian.net/browse/SQONE-521
[SQONE-519]: https://moderntribe.atlassian.net/browse/SQONE-519

### DIFF
--- a/wp-content/themes/core/components/navigation/Navigation_Controller.php
+++ b/wp-content/themes/core/components/navigation/Navigation_Controller.php
@@ -1,0 +1,81 @@
+<?php declare( strict_types=1 );
+
+namespace Tribe\Project\Templates\Components\navigation;
+
+use Tribe\Libs\Utils\Markup_Utils;
+use Tribe\Project\Nav_Menus\Menu;
+use Tribe\Project\Templates\Components\Abstract_Controller;
+
+class Navigation_Controller extends Abstract_Controller {
+
+	public const LOCATION         = 'location';
+	public const CLASSES          = 'classes';
+	public const ATTRS            = 'attrs';
+	public const NAV_LIST_CLASSES = 'nav_list_classes';
+	public const NAV_MENU_ARGS    = 'nav_menu_args';
+
+	private const DEFAULT_NAV_MENU_ARGS = [
+		'container'   => false,
+		'menu_class'  => '',
+		'menu_id'     => '',
+		'depth'       => 1,
+		'items_wrap'  => '%3$s',
+		'fallback_cb' => false,
+	];
+
+	private string $location;
+	private array $classes;
+	private array $attrs;
+	private array $nav_list_classes;
+	private array $nav_menu_args;
+
+	public function __construct( array $args = [] ) {
+		$args = $this->parse_args( $args );
+
+		$this->location         = (string) $args[ self::LOCATION ];
+		$this->classes          = (array) $args[ self::CLASSES ];
+		$this->attrs            = (array) $args[ self::ATTRS ];
+		$this->nav_list_classes = (array) $args[ self::NAV_LIST_CLASSES ];
+		$this->nav_menu_args    = $this->parse_menu_args( $args[ self::NAV_MENU_ARGS ] );
+	}
+
+	protected function defaults(): array {
+		return [
+			self::LOCATION         => '',
+			self::CLASSES          => [ 'c-nav' ],
+			self::ATTRS            => [],
+			self::NAV_LIST_CLASSES => [ 'c-nav__list' ],
+			self::NAV_MENU_ARGS    => [],
+		];
+	}
+
+	private function parse_menu_args( $menu_args ): array {
+		$menu_args['theme_location'] = $this->location;
+
+		return wp_parse_args( $menu_args, self::DEFAULT_NAV_MENU_ARGS );
+	}
+
+	public function has_menu(): bool {
+		return has_nav_menu( $this->location );
+	}
+
+	public function get_classes(): string {
+		return Markup_Utils::class_attribute( $this->classes );
+	}
+
+	public function get_attrs(): string {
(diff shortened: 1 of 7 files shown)


wp-content/themes/core/components/navigation/navigation.php
```diff
@@ -0,0 +1,21 @@
...
+// phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UndefinedVariable
+$c = Navigation_Controller::factory( $args );
+
+if ( ! $c->has_menu() ) {
+	return;
+}
+
+?>
+<nav <?php echo $c->get_classes(); ?> <?php echo $c->get_attrs(); ?>>
+	<ol <?php echo $c->get_nav_list_classes(); ?>>
```
ScottGarrison, line 17: This is probably picky but I feel like OL works better for navigation such as a breadcrumb where we don't know the use case of the navigation here.

ScottGarrison (approved):
:+1:

wp-content/themes/core/components/navigation/Navigation_Controller.php
```diff
@@ -0,0 +1,82 @@
...
+		'container'   => false,
+		'menu_class'  => '',
+		'menu_id'     => '',
+		'depth'       => 1,
+		'items_wrap'  => '%3$s',
+		'fallback_cb' => false,
+	];
+
+	private string $location;
+	private array $classes;
```
defunctl, line 27: If you wanted to save me some time when moving to the PHPCS v2 rules you can add the docblock for the type that each array is going to contain, so some IDE's will warn the user when using the using incorrect data, e.g.

```php
	/**
	 * @var string[]
	 */
```

wp-content/themes/core/components/navigation/Navigation_Controller.php
```diff
@@ -0,0 +1,82 @@
...
+		$args = $this->parse_args( $args );
+
+		$this->location         = (string) $args[ self::LOCATION ];
+		$this->classes          = (array) $args[ self::CLASSES ];
+		$this->attrs            = (array) $args[ self::ATTRS ];
+		$this->nav_list_classes = (array) $args[ self::NAV_LIST_CLASSES ];
+		$this->nav_menu_args    = $this->parse_menu_args( $args[ self::NAV_MENU_ARGS ] );
+	}
+
+	protected function defaults(): array {
```
defunctl, line 42: This will be fixed automatically with `phpcbf` in the v2 rules, but class order should go public methods, protected and then private.

wp-content/themes/core/components/navigation/Navigation_Controller.php
```diff
@@ -0,0 +1,82 @@
...
+		return [
+			self::LOCATION         => '',
+			self::CLASSES          => [ 'c-nav' ],
+			self::ATTRS            => [],
+			self::NAV_LIST_CLASSES => [ 'c-nav__list' ],
+			self::NAV_MENU_ARGS    => [],
+		];
+	}
+
+	private function parse_menu_args( $menu_args ): array {
```
defunctl, line 52: missing type

wp-content/themes/core/components/navigation/Navigation_Controller.php
dpellenwood: Sounds good. Related: for a key => value array (such as attributes), is there a proper style for hinting the type required for both the key and the value? (As supposed to the `string[]` where just the value type is hinted.)

dpellenwood:
Thanks Justin. I'll make these changes with my next PR (for the site footer).

wp-content/themes/core/components/navigation/Navigation_Controller.php
defunctl: You bet! Here are some examples

-   `int[]`: an array with only INT values -- [1, 4, 6, 8, 9, ...]
-   `array<int, int>`: an array with only INT values -- [4 => 1, 8 => 4, 12 => 6, ...]
-   `string[]`: an array with only STRING values -- [„foo", „bar", ...]
-   `array<int, string>`: an array with only STRING values -- [4 => „foo", 8 => „bar", ...]
-   `Order[]`: an array with only „Order"-Object values -- [Order, Order, ...]
-   `array<int|string, Order>`: an array with INT or STRING as key and „Order"-Object values -- [4 => Order, 'foo' => Order, ...]
-   `array<int|string, mixed>`: an array with INT or STRING as key and mixed as values -- [1 => 1, 4 => „foo", 6 => \stdClass, ...]
-   `array<int, array<int, string>>`: an array with INT as key and and an array (with INT as key and string as value) as values -- [1 => [1 => „foo"], 4 => [1 => 4], ...]
-   `array<int, string[]>`: an array with INT as key and and an array (with INT as key and string as value) as values -- [1 => [„foo", „lall"], 4 => [„öäü", „bar"], ...]
-   `array{output: string, debug: string}`: an array with the key „output" and „debug", the values are STRING values -- [‚output' => ‚foo', ‚debug' => ‚bar']
-   `array<int, array{output: string, debug: string}>`: an array with the key „output" and „debug", the values are STRING values -- [1 => [‚output' => ‚foo', ‚debug' => ‚bar'], 3 => [‚output' => ‚foo', ‚debug' => ‚bar'], ...]

[sauce](https://dev.to/suckup_de/modern-phpdoc-annotations-11d4)

Or some other basic associative arrays:

- `string[][]`
- `int[][]`

etc...

PR #872, opened 2021-10-06
React apps

jorgemasta:
## What does this do/fix?

After upgrading to webpack 5, the dev mode for react apps was broken. I have updated the configuration to get it working again.

Next steps:
- Enable HTTPS in dev mode. Seems quite complex as you have to generate and sign certificates on each device.
- Enable HMR (Hot Module Replacement) with Live Reload. In theory it is configured but it's not working.
- Update example app reducing its complexity (bye redux/sagas). Include a basic example with [`@moderntribe/faceted-loops`](https://github.com/moderntribe/faceted-loops)?

## Tests

Does this have tests?

No, this doesn't need tests because I have changed only some webpack settings

### DIFF
--- a/package.json
+++ b/package.json
@@ -130,6 +130,7 @@
     "babel-loader": "^8.0.6",
     "babel-plugin-lodash": "^3.3.4",
     "babel-plugin-module-resolver": "^3.2.0",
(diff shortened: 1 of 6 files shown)


jorgemasta:
Closed in favor of #873 (gitflow)

PR #873, opened 2021-10-06
Support react apps in dev mode

jorgemasta:
## What does this do/fix?

After upgrading to webpack 5, the dev mode for react apps was broken. I have updated the configuration to get it working again.

Next steps:
- Enable HTTPS in dev mode. Seems quite complex as you have to generate and sign certificates on each device.
- Enable HMR (Hot Module Replacement) with Live Reload. In theory it is configured but it's not working.
- Update example app reducing its complexity (bye redux/sagas). Include a basic example with [`@moderntribe/faceted-loops`](https://github.com/moderntribe/faceted-loops)?

Other ideas:
- Try to use `@wordpress/scripts` to run our react apps
- Integrate react apps as gutenberg blocks

## Tests

Does this have tests?

No, this doesn't need tests because I have changed only some webpack settings

### DIFF
--- a/package.json
+++ b/package.json
@@ -130,6 +130,7 @@
     "babel-loader": "^8.0.6",
     "babel-plugin-lodash": "^3.3.4",
     "babel-plugin-module-resolver": "^3.2.0",
+    "babel-plugin-react-css-modules": "^5.2.6",
     "browser-sync": "^2.26.7",
     "concat": "^1.0.3",
     "core-js": "^3.13.1",
--- a/webpack/configs/app-base.js
+++ b/webpack/configs/app-base.js
@@ -1,26 +1,48 @@
 /**
  * External Dependencies
  */
+const path = require( 'path' );
 const webpack = require( 'webpack' );
 
 module.exports = {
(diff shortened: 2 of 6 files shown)


nickpelton:
Dropping everything to Http is problematic as we generate https certs for our local dev enviroments already... is it possible ot just use browsersync and it's proxy or is this something different? 

Also, I love the idea of updating the example app to drop ducks/sagas and switch to something more standard like react hooks or another main stream state manager. I assume this will be on another PR though? Also, Wordpress is building it's own script management, worth looking at that as an alternative or should we stick when what we got?

nickpelton:
Hey Jorge, pending merge here. Is this ready to go?

jorgemasta:
> Hey Jorge, pending merge here. Is this ready to go?

Yes! It's ready. Tomorrow morning I will check that merge conflicts

jorgemasta:
> Hey Jorge, pending merge here. Is this ready to go?

Yes. It's ready

nickpelton:
Go ahead and Merge!

PR #881, opened 2021-11-19
Bugfix/jira ticket prefixing

defunctl:
## What does this do/fix?

- Fixes bad prefixed commits when using sprint or release branches that contain numbers after them.

## QA

- You should not get any prefixing when committing a branch like `release/21.11.01` or `sprint/21.01`
- Branches such as `feature/XXX-21/my-feature` should still properly get prefixed with `[XXX-21]: original commit...`

## Tests

Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests because it's a lefthook update.


nickpelton (approved):
Regex is magic 🪄

Issue #880, opened 2021-11-18
Improve filter on nav menu classes


crstauf:
(Sorry if this should be a discussion first.)

Because menus can be loaded via the `menu` (menu ID, slug, name, or object) parameter of `wp_nav_menu()`, this function should be adjusted:

1. if the classes are relevant only for menus includes via `theme_location`, check if `theme_location` is populated before adding classes
1. adjust to support menus included via `menu` parameter and not `theme_location`

Happy to submit a PR once a decision is made.

https://github.com/moderntribe/square-one/blob/6ae17da9de0fc3f64e37e370b651d3efc2eadbc3/wp-content/plugins/core/src/Nav_Menus/Nav_Attribute_Filters.php#L44-L85

nickpelton:
Hey @crstauf This looks like a worthy change. We typically start with a PR rather than issues, as we track most tasks in Jira. I'm going to close this, looking forward to reviewing the PR.

crstauf:
@nickpelton Which option should I proceed with: 1 or 2?

crstauf:
Discussed in Slack: supporting adding classes for both `theme_location` and `menu` parameters.

PR #883, opened 2021-11-30
Apply nav menu classes only if `theme_location` parameter is populated

crstauf:
## What does this do/fix?

Improves the classes added to nav menu elements by supporting the `menu` parameter (of `wp_nav_menu()`). See #880.

## Tests

Does this have tests?

- [ ] Yes

After creating two menus (`sample` and `primary`) and assigning one of them to the `primary` theme location, added the following code to an mu-plugin to easily test the proper output:

```php
<?php declare(strict_types=1);

use Tribe\Project\Nav_Menus\Menu;

add_action( 'wp_body_open', function() {

	// Test theme_location parameter.
	echo Menu::menu( [
		'theme_location' => 'primary',
		'container_class' => 'test-1 location-primary',
	] );

	// Test menu parameter.
	echo Menu::menu( [
		'menu' => 'sample',
		'container_class' => 'test-2 menu-sample',
	] );

	// Test menu and theme_location parameter.
	echo Menu::menu( [
		'menu' => 'sample',
		'theme_location' => 'primary',
		'container_class' => 'test-3 menu-sample',
	] );

	// Test invalid menu and valid theme_location parameter.
	echo Menu::menu( [
		'menu' => 'foo',
		'theme_location' => 'primary',
		'container_class' => 'test-4 location-primary',
	] );

	// Test valid menu and invalid theme_location parameter.
	echo Menu::menu( [
		'menu' => 'sample',
		'theme_location' => 'bar',
		'container_class' => 'test-5 menu-sample',
	] );

	exit;
} );
```

### DIFF
--- a/wp-content/plugins/core/src/Nav_Menus/Nav_Attribute_Filters.php
+++ b/wp-content/plugins/core/src/Nav_Menus/Nav_Attribute_Filters.php
@@ -42,6 +42,10 @@ class Nav_Attribute_Filters {
 	 * @filter nav_menu_css_class
 	 */
 	public function customize_nav_item_classes( array $classes, object $item, stdClass $args, int $depth ): array {
+		if ( empty( $args->theme_location ) ) {
+			return $classes;
+		}
+
 		$theme_location = $args->theme_location;
 
 		$classes[] = $theme_location . '__list-item';
@@ -105,6 +109,10 @@ class Nav_Attribute_Filters {
 	 * @filter nav_menu_link_attributes
 	 */
 	public function customize_nav_item_anchor_atts( array $atts, object $item, stdClass $args, int $depth ): array {
+		if ( empty( $args->theme_location ) ) {
+			return $atts;
+		}
+
 		$theme_location = $args->theme_location;
 
 		$classes = [
--- a/wp-content/plugins/core/src/Nav_Menus/Nav_Attribute_Filters.php
+++ b/wp-content/plugins/core/src/Nav_Menus/Nav_Attribute_Filters.php
@@ -12,6 +12,61 @@ use stdClass;
 class Nav_Attribute_Filters {
 
 	/**
+	 * Generate a class prefix based on wp_nav_menu() parameters.
+	 *
+	 * Because `menu` overrides `theme_location`, and `theme_location` is used if
+	 * `menu` is empty or menu does not exist, the solution is to get the slugs
+	 * of both menus, and if they're equal, use the theme location. If they're not equal,
+	 * then use the menu slug.
+	 *
+	 * @param stdClass $args An object of {@see wp_nav_menu()} arguments.
+	 *
+	 * @return string
+	 *
+	 * @todo evaluate: prepending "location-" and "menu-" is backwards-compatibility issue
+	 */
+	protected function get_class_prefix( stdClass $args ): string {
+		$menu = wp_get_nav_menu_object( $args->menu );
+		$locations = get_nav_menu_locations();
+
+		$menu_slug = '';
+		$location_menu_slug = '';
+
+		// Get slug of menu if exists.
+		if ( $menu ) {
+			$menu_slug = $menu->slug;
+		}
+
+		// Get theme location's menu.
+		if ( $args->theme_location && $locations && isset( $locations[ $args->theme_location ] ) ) {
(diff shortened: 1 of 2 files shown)


wp-content/plugins/core/src/Nav_Menus/Nav_Attribute_Filters.php
```diff
@@ -11,6 +11,61 @@
...
+	 * then use the menu slug.
+	 *
+	 * @param stdClass $args An object of {@see wp_nav_menu()} arguments.
+	 *
+	 * @return string
+	 *
+	 * @todo evaluate: prepending "location-" and "menu-" is backwards-compatibility issue
+	 */
+	protected function get_class_prefix( stdClass $args ): string {
+		$menu = wp_get_nav_menu_object( $args->menu );
```
defunctl, line 29: This should fail PHPCS

wp-content/plugins/core/src/Nav_Menus/Nav_Attribute_Filters.php
```diff
@@ -11,6 +11,61 @@
...
+		$locations = get_nav_menu_locations();
+
+		$menu_slug = '';
```
defunctl, line 32: So should this :)

wp-content/plugins/core/src/Nav_Menus/Nav_Attribute_Filters.php
```diff
@@ -11,6 +11,61 @@
...
+		$location_menu_slug = '';
+
+		// Get slug of menu if exists.
```
defunctl, line 35:
```suggestion
		// Get slug of menu if it exists.
```

wp-content/plugins/core/src/Nav_Menus/Nav_Attribute_Filters.php
```diff
@@ -11,6 +11,61 @@
...
+		if ( $menu ) {
```
defunctl, line 36: `wp_get_nav_menu_object()` is filterable, I'd check directly that `$menu->slug` itself exists, and as a suggestion you could combine that with the initial define which improves readability. 

At some point scrolling through this large code block you have to back up and check if things have been initialized or not.

e.g.

```php
$menu_slug = ( $menu->slug  ?? '' ) ? $menu->slug : '';
```

defunctl (changes requested):
This would actually be a really great use case for https://github.com/moderntribe/square1-pipeline. 

Rather than a large method with lots of "if" statements, you can move the logic into defined steps/stages and skip to the next step if a condition or two isn't met. 

That would also make each step object easily testable with integration tests, where right now this would be very difficult to write tests for, given the number of conditions in a single method.

dpellenwood (commented):
@crstauf thank you for taking a stab at this. However, it actually leads to more issues than it solves. I'd like to chat with you in person about it before we proceed with another approach.

wp-content/plugins/core/src/Nav_Menus/Nav_Attribute_Filters.php
crstauf: > check directly that $menu->slug itself exists

👍 

> At some point scrolling through this large code block you have to back up and check if things have been initialized or not.

I very rarely use ternary operators, as it decreases readability, especially for junior devs. Remembering the variables were initialized or scrolling up a little bit to confirm is much easier. ✌️

dpellenwood (changes requested):
Thanks again for chatting through this with me, @crstauf . Per our conversation, please take the approach of checking for a theme location and if there is not one provided, existing early from the class filters. We don't (ever) want to create CSS classes for menus & items based on the menu ID/slug because those are specific to the data of the menu, not generic to a menu's template/usage/style.

nickpelton:
Thanks everyone for popping on and chatting this through, good discussion leads to better tools for us all.

dpellenwood (approved):
:rocket: Thanks again for being patient while we all hashed this out together!

nickpelton (approved):

defunctl:
@crstauf don't forget a changelog update, please

crstauf:
@defunctl Waiting for your approval.

defunctl (approved):
🚀

PR #887, opened 2021-12-09
Feature/docker compose v2

defunctl:
## What does this do/fix?

- Removes the `WARN[0000] network proxy: network.external.name is deprecated in favor of network.name` when using `docker compose` (v2).

**Note:** existing projects should update their `docker-compose.yml` files as well, point to this PR for you respective BE devs.

## QA

- docker-compose v1 should work as it always did.
- docker compose v2 support will require a new version of `so`: https://github.com/moderntribe/square1-global-docker/pull/94

## Tests

Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests because it's a docker-compose.yml change.


nickpelton (approved):
Tested on related PR: https://github.com/moderntribe/square1-global-docker/pull/94#issuecomment-990286489

PR #888, opened 2021-12-10
SQONE-523 Section Nav

dpellenwood:
## What does this do/fix?
Builds out the Section Nav component for use in SquareOne. Note that this is the base component work. The Section Nav Block is a separate branch with an (incoming) PR.

[SQONE-523]

[SQONE-523]: https://moderntribe.atlassian.net/browse/SQONE-523?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wp-content/themes/core/components/navigation/Navigation_Controller.php
```diff
@@ -8,75 +8,122 @@
...
+		'depth'          => 1,
+		'items_wrap'     => '%3$s',
+		'fallback_cb'    => false,
+		'theme_location' => '',
 	];
 
-	private string $location;
-	private array $classes;
+	/**
+	 * @var array
```
nickpelton, line 31: 🟠 
Should fix: Since we are declaring the types, we don't need the var comments anymore.

wp-content/themes/core/components/section_nav/Section_Nav_Controller.php
```diff
@@ -0,0 +1,220 @@
...
+	 */
+	private array $container_classes;
+
+	/**
+	 * @var int|string|\WP_Term
+	 */
+	private $menu;
+
+	/**
+	 * @var string
```
nickpelton, line 51: 🟠  Should fix: If it's a single value like this, the type is all that is required.

wp-content/themes/core/components/section_nav/Section_Nav_Controller.php
```diff
@@ -0,0 +1,220 @@
...
+	 */
+	private string $container_id;
+
+	/**
+	 * @var string
+	 */
+	private string $more_id;
+
+	public function __construct( array $args = [] ) {
+		$this->unique_id    = uniqid();
```
nickpelton, line 91: ❓ 
Curious what we need a UUID for?

wp-content/themes/core/components/section_nav/css/section-nav.pcss
```diff
@@ -0,0 +1,218 @@
...
+.c-section-nav {
+	border: 1px solid var(--color-border);
+
+	@media (--viewport-small) {
+		overflow: hidden;
+		border: none;
+	}
+
+	/* CASE: the more menu is in use. */
+	&.c-section-nav--more-initialized {
```
nickpelton, line 16: 💡 
I've see this pattern al throughout our code. Any reason we don't include this in the naming like:

`&.c-section-nav--more-case-initialized`

I've see other patterns like this that say things like `js` in the name so you know this modifier is managed by something else. Seeing that `case` would be a flag that this is changed by some logic somewhere based on a case.

wp-content/themes/core/components/section_nav/js/section-nav.js
```diff
@@ -0,0 +1,360 @@
...
+
+/**
+ * Build events for this module.
+ */
+const bindEvents = () => {
+	delegate( el.container, '[data-js="c-section-nav__toggle--more"]', 'click', toggleMoreMenu );
+	delegate( el.container, '[data-js="c-section-nav__toggle--more"]', 'keyup', handleEscKeyUp );
+	delegate( el.container, '[data-js="c-section-nav__toggle--mobile"]', 'click', toggleSectionNav );
+	delegate( el.container, '[data-js="c-section-nav__toggle--mobile"]', 'keyup', handleEscKeyUp );
+	delegate( el.container, '.section-nav__list-item a', 'keyup', handleEscKeyUp );
```
nickpelton, line 337: ❓ 
What's the diff between delegate and on here?

wp-content/themes/core/components/section_nav/js/section-nav.js
```diff
@@ -0,0 +1,360 @@
...
+
+	on( document, 'click', handleClickOut );
+	on( document, 'keyup', handleTabKeyUp );
+	on( document, 'modern_tribe/resize_executed', handleResize );
```
nickpelton, line 341: ❓ 
What is `modern_tribe/resize_executed`?

wp-content/themes/core/components/section_nav/js/section-nav.js
```diff
@@ -0,0 +1,360 @@
...
+
+	el.sectionNavs = sectionNavs;
+
+	handleInitialState();
+	bindEvents();
+
+	console.info( 'SquareOne Theme: Initialized Section Nav component scripts.' );
+};
+
+export default init;
```
nickpelton, line 360: 🌮 😅 Nav management is fun isn't it. That's a lot of code. Nice work!

wp-content/themes/core/components/section_nav/section_nav.php
```diff
@@ -0,0 +1,33 @@
...
+<div <?php echo $c->get_classes(); ?> <?php echo $c->get_attrs(); ?>>
+
+	<?php echo $c->get_mobile_toggle(); ?>
+
+	<div <?php echo $c->get_container_classes(); ?> <?php echo $c->get_container_attrs(); ?>>
+		<?php echo $c->get_desktop_label(); ?>
+		<?php echo $c->get_nav_menu(); ?>
+	</div>
+
+	<template data-template="more">
```
nickpelton, line 21: 🌮  Nice use of a JS template here.

nickpelton (approved):
Looking Great! Mostly questions for myself as I learn our FE patterns. 

How would you like to QA test this?

wp-content/themes/core/components/section_nav/Section_Nav_Controller.php
dpellenwood: Good question. For the ARIA attributes required to associate a button with the thing it controls, so the mobile show/hide button and the container that is shown/hidden; and [the "More" button](https://github.com/moderntribe/square-one/pull/888/files#diff-830a1fc2c6e6373e01e4eb1df77af46ae9a258e201b2819d5d144dfe8df4cc01R152) for showing/hiding whatever menu items are tucked into [the drawer container on desktop](https://github.com/moderntribe/square-one/pull/888/files#diff-830a1fc2c6e6373e01e4eb1df77af46ae9a258e201b2819d5d144dfe8df4cc01R208).

Since the ARIA association works with the `id` attribute, we need a unique "id" for each section nav on a page.

wp-content/themes/core/components/section_nav/css/section-nav.pcss
dpellenwood: The [BEM syntax itself](http://getbem.com/introduction/) (the double-dash) is intended to inform you that this is a conditional case or a "modifier". We just try to add CSS comments whenever we use them to to clarify in human-readable format that the conditional modifier is.

wp-content/themes/core/components/section_nav/js/section-nav.js
dpellenwood: the `delegate()` helper basically watches the container element and assigns the relevant event callbacks to the selector(s) any time they exist (or are added to) the document. Whereas, `on()` events will only be applied when the document is loaded. So, delegate event hooks are useful if the selector you want to add doesn't exist in the DOM when the document's load/ready methods are fired.

wp-content/themes/core/components/section_nav/js/section-nav.js
dpellenwood: It's an [existing](https://github.com/moderntribe/square-one/blob/main/wp-content/themes/core/assets/js/src/theme/core/resize.js#L15) [(already denounced) document resize event handler](https://github.com/moderntribe/square-one/blob/main/wp-content/themes/core/assets/js/src/theme/core/ready.js#L39) we used throughout Sq1.

dpellenwood (commented):
Thanks, Nick. I'll get those property type comments removed. Let me know if any of my comment responses aren't clear enough.

wp-content/themes/core/components/section_nav/Section_Nav_Controller.php
nickpelton: Makes sense, it  does feel a-little inefficient to generate a completely new UUID for every execution though. Unless there is another way to do an ID key on the component, when there will possibly be multiple of them on a page, let's leave it as-is.

wp-content/themes/core/components/section_nav/js/section-nav.js
nickpelton: Thanks!

wp-content/themes/core/components/section_nav/js/section-nav.js
nickpelton: #TIL

wp-content/themes/core/components/section_nav/css/section-nav.pcss
nickpelton: Totally makes sense. I'm just thinking in cases of "What is managing this class". Maybe it's not important, but in cases where we're adding/removing classes with PHP/JS it certainly can be tricky to reverse engineer things.

wp-content/themes/core/components/section_nav/Section_Nav_Controller.php
ChrisMKindred: [ACF provides a unique identifier](https://www.advancedcustomfields.com/resources/blocks/) for each block, but it looks like we aren't passing it out of Base_Model. I think we should add `Id` to the `name`, `classes`, `data`, and `mode` there, and then we won't need this step and could just call `$this->id` for anything that requires it.

ChrisMKindred (approved):
This has made me want to get back into JS more... I added a comment to answer a question Nick had. I am happy to implement it for us if it would work better than needing to use the uniquid() function.

wp-content/themes/core/components/section_nav/Section_Nav_Controller.php
dpellenwood: Thanks @ChrisMKindred , while it's true, the ACF block ID will work for instances which are defined via a block, the component could just as easily be used as a direct template part in a single page or post template where we wouldn't have access to a block ID. I suppose we could modify the component to accept a unique ID of some sort and then pass along the block ID or require it when defining the usage outside that context. But, it seems just as easy to allow the each component to set their own UUID at the time of usage.  We use [this same `uniqid()` approach pretty regularly throughout SquareOne](https://github.com/moderntribe/square-one/search?q=uniqid%28) & the projects we've built previously.

PR #889, opened 2021-12-10
SQONE-668 Section Nav Block

dpellenwood:
## What does this do/fix?
Builds out the Section Nav Block based on the Section Nav Component (PR #888 ).

[SQONE-668]

[SQONE-668]: https://moderntribe.atlassian.net/browse/SQONE-668?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wp-content/themes/core/blocks/sectionnav/sectionnav.php
```diff
@@ -0,0 +1,8 @@
+<?php declare(strict_types=1);
+
+/**
+ * @var array $args Arguments passed to the template
+ */
+// phpcs:ignore VariableAnalysis.CodeAnalysis.VariableAnalysis.UndefinedVariable
+$model = new \Tribe\Project\Blocks\Types\Section_Nav\Section_Nav_Model( $args['block'] );
+get_template_part( 'components/blocks/section_nav/section_nav', null, $model->get_data() );
```
nickpelton, line 8: Since this is literally a "block" placeholder for the section nav component, do we need this? Further down all we're doing is farther down is passing this same data to the section_nav component. Seems like an unnecessary shell.

wp-content/themes/core/components/blocks/section_nav/section_nav.php
```diff
@@ -0,0 +1,22 @@
...
+$c = Section_Nav_Block_Controller::factory( $args );
+
+/**
+ * This block is nothing more than a data source for the Section Nav component.
+ *
+ * In order for the sticky-navigation feature to work properly, the section nav component
+ * cannot be wrapped in a container element. This is a limitation of the `display: sticky`
+ * styling used to create the sticky-on-scroll effect.
+ */
+get_template_part(
```
nickpelton, line 18: Here is that shell I mentioned.

wp-content/themes/core/components/section_nav/Section_Nav_Controller.php
```diff
@@ -103,11 +104,11 @@ public function __construct( array $args = [] ) {
 		$this->desktop_label     = (string) $args[ self::DESKTOP_LABEL ];
 		$this->more_label        = (string) $args[ self::MORE_LABEL ];
 		$this->sticky            = (bool) $args[ self::STICKY ];
-		$this->mobile_init_open  = (bool) $args[ self::MOBILE_INIT_OPEN ];
+		$this->mobile_init       = (string) $args[ self::MOBILE_INIT ];
```
nickpelton, line 107: Any reason we can't keep it a bool? Checking a 'true'/'false' string is not ideal.

wp-content/themes/core/components/section_nav/js/section-nav.js
```diff
@@ -218,36 +228,74 @@ const handleNavFit = ( sectionNav ) => {
...
+ * Set up the IntersectionObserver callback for each SectionNav.
+ *
+ * @type {IntersectionObserver}
+ */
+const sectionNavIntersectionObserver = new IntersectionObserver( ( [ entry ] ) => {
+	entry.target.classList.toggle( 'c-section-nav--stuck', entry.intersectionRatio < 1 );
+}, intObserverOpts );
+
+/**
+ * Initialize the observers applied to each SectionNav.
```
nickpelton, line 247:
```suggestion
 * Initialize the observers to the SectionNav.
```

nickpelton (approved):
🚀 
Looking awesome. Some minor suggestions.

ChrisMKindred (approved):
Looks good to me.

wp-content/themes/core/blocks/sectionnav/sectionnav.php
dpellenwood: Hmmm... I suppose we _could_. I'm trying to consider that in the future a design or scope decision may require adding additional fields/settings/values to the _block_ which could then be different than the base component. But, that may be overkill, too. I do like the separation between the content-entered "block" and the generic template part "component" though. Lemme stew on this over the weekend.

wp-content/themes/core/components/section_nav/Section_Nav_Controller.php
dpellenwood: The block registration is using a button_group for the options (Closed/Open), and that ACF field requires values I suppose I could reduce the return value from the Block's Model to be a boolean, rather than the actual ACF value. I'll update that per your suggestion.

wp-content/themes/core/blocks/sectionnav/sectionnav.php
nickpelton: 👍  I don't consider this a blocking issue as it does have some utility as mentioned.

PR #890, opened 2021-12-13
Improves and fixes automated script preloading

crstauf:
## What does this do/fix?

- prevents scripts with no source and no dependencies from being preloaded, resulting in a duplicate and redundant request of the current document
- handles preloading of dependencies of asset with no source url

**Discovery**
If `jquery` (an alias of `jquery-core`) is enqueued in the footer, it has no source, and so the current page URL is preloaded, resulting in a duplicate and unused preload of the document. Initially I prevented preloading of dependency aliases, but then decided to support preloading of those dependencies.

See moderntribe/data-org#412. Issue can be currently seen on https://data.org/news/: look for `?ver=3.4.1` in _Name_ column of _Network_ tab in DevTools and a warning in _Console_ tab.


dpellenwood (approved):
Thanks for tackling this, Caleb. :scroll: :dash:

nickpelton (approved):
Lovely improvement. Thanks!

PR #902, opened 2021-12-31
Feature/coding standards v2

defunctl:
## What does this do/fix?

Brings SquareOne up to date with [coding standards v2](https://github.com/moderntribe/coding-standards/tree/2.0.x).

Sorry, Last massive PR of 2021 :partying_face: 

This should get reviewed/merged ASAP please, otherwise we're going to have ongoing merge conflict nightmares.

This is 98% PHPCS fixes and 2% bug fixes/updates as I came across them while converting the framework.


wp-content/plugins/core/src/Routes/Routes_Definer.php
```diff
@@ -1,11 +1,5 @@
 <?php declare(strict_types=1);
 
-/**
```
nickpelton, line 3: Why remove contextual comments?

wp-content/plugins/core/src/Routes/Routes_Subscriber.php
```diff
@@ -1,11 +1,5 @@
 <?php declare(strict_types=1);
 
-/**
```
nickpelton, line 3: Why remove contextual comments?

wp-content/plugins/core/src/Routes/Sample_REST_Route.php
```diff
@@ -1,14 +1,11 @@
 <?php declare(strict_types=1);
 
-/**
```
nickpelton, line 3: Why remove contextual comments?

wp-content/plugins/core/src/Routes/Sample_Route.php
```diff
@@ -1,12 +1,5 @@
 <?php declare(strict_types=1);
 
-/**
```
nickpelton, line 3: Why remove contextual comments?

wp-content/plugins/core/src/Templates/Components/Abstract_Controller.php
```diff
@@ -13,6 +13,15 @@
  */
 abstract class Abstract_Controller {
 
+	/**
+	 * @param array $args
+	 *
+	 * @return static
+	 */
+	public static function factory( array $args = [] ) {
```
nickpelton, line 21: Looks like we don't get `:static` till PHP8. 😭  https://wiki.php.net/rfc/static_return_type

wp-content/plugins/core/src/Theme/Branding/Login_Screen.php
```diff
@@ -27,7 +27,7 @@ public function inject_login_logo(): void {
 		$logo_height = $logo_data[2] / 2;
 
 		?>
-		<style type="text/css">
+		<style>
```
nickpelton, line 30: Why remove the type?

wp-content/plugins/core/src/Theme/Config/Image_Sizes.php
```diff
@@ -19,7 +19,10 @@ class Image_Sizes {
 	public const SOCIAL_SHARE_OPENGRAPH = 'social-share-opengraph';
 	public const SOCIAL_SHARE_TWITTER   = 'social-share-twitter';
 
-	private $sizes = [
+	/**
+	 * @var array<string, mixed>
```
nickpelton, line 23: This comment format differs from previous where I've seen `[]string|mixed`

wp-content/plugins/core/src/Theme/Config/Web_Fonts.php
```diff
@@ -2,15 +2,18 @@
...
-	public const  PROVIDER_TYPEKIT = 'typekit';
-	public const  PROVIDER_GOOGLE  = 'google';
 	public const  PROVIDER_CUSTOM  = 'custom';
+	public const  PROVIDER_GOOGLE  = 'google';
+	public const  PROVIDER_TYPEKIT = 'typekit';
 	private const TYPEKIT_API      = 'https://use.typekit.net';
 
 	/**
-	 * @var array
+	 * @var array<string, string>
```
nickpelton, line 16: Same as previous comment. This is different than other comments I've seen.

wp-content/plugins/core/src/Theme/Media/Image_Wrap.php
```diff
@@ -7,19 +7,19 @@ class Image_Wrap {
...
 	 */
-	public function customize_wp_image_non_captioned_output( $html ) {
+	public function customize_wp_image_non_captioned_output( string $html ): string {
 
 		if ( ! is_singular() && ! in_the_loop() && ! is_main_query() ) {
 			return $html;
 		}
 
-		return preg_replace_callback( '/<p>((?:.(?!p>))*?)(<a[^>]*>)?\s*(<img[^>]+>)(<\/a>)?(.*?)<\/p>/i', static function ( $matches ) {
+		$result = preg_replace_callback( '/<p>((?:.(?!p>))*?)(<a[^>]*>)?\s*(<img[^>]+>)(<\/a>)?(.*?)<\/p>/i', static function ( $matches ) {
```
nickpelton, line 22: This seems like a pretty big bug. Might need to look into the implications here.

wp-content/plugins/core/src/Theme/Media/Oembed_Filter.php
```diff
@@ -120,17 +136,14 @@ protected function get_fresh_frontend_html( $html, $url, $attr ) {
...
-	 *
-	 * @filter embed_oembed_html 99
-	 */
-	public function wrap_admin_oembed( $html, $url, $attr, $post_id ) {
-		if ( ! is_admin() ) {
-			return $html;
-		}
 
-		return sprintf( '<div class="wp-embed"><div class="wp-embed-wrap">%s</div></div>', $html );
+	private function get_layout_container_attrs( string $provider_name, string $embed_id, string $title ): array {
```
nickpelton, line 140: Why move?

nickpelton (commented):
Some general comments. Got about half way through and wasn't seeing anything new.

wp-content/plugins/core/functions/panels.php
```diff
@@ -9,7 +9,7 @@
  * @since tribe-square-one 1.0
```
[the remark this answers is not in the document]
defunctl: I think that's a good idea, I didn't want to actually delete anything at the time. I

wp-content/plugins/core/src/Assets/Theme/Scripts.php
```diff
@@ -139,6 +131,14 @@ public function enqueue_scripts(): void {
 		wp_enqueue_script( 'comment-reply' );
 	}
 
+	protected function print_preload_tag( _WP_Dependency $script ): void {
```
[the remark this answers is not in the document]
defunctl: Yep, phpcs does it automatically. As a developer when looking at a class, we really care about the interface, the thing we can use (unless we're actually developing/modifying the feature), so this enforces public methods first, so if I was going to just use some class, its interface is clearer.

https://github.com/slevomat/coding-standard#slevomatcodingstandardclassesclassstructure-

wp-content/plugins/core/src/Blocks/Blocks_Definer.php
```diff
@@ -114,7 +113,7 @@ public function define(): array {
 				} ),
 			] ),
 
-			Block_Deny_List::class => DI\create()->constructor( DI\get( self::DENY_LIST ) ),
+			Block_Deny_List::class => DI\autowire()->constructor( DI\get( self::DENY_LIST ) ),
```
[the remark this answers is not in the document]
defunctl: Say I needed to modify `Block_Deny_List`, and I need to inject an instance of something else and it was a concrete object, because it's set to autowire now, I don't have to change anything in the definer at all.

wp-content/plugins/core/src/Theme/Branding/Login_Screen.php
defunctl: phpstorm told me it is deprecated :man_shrugging:

wp-content/plugins/core/src/Integrations/Google_Tag_Manager/GTM_Scripts.php
```diff
@@ -6,10 +6,7 @@
 
 class GTM_Scripts {
 
-	/**
-	 * @var \Tribe\Project\Object_Meta\Analytics_Settings
-	 */
-	private $settings;
+	private \Tribe\Project\Object_Meta\Analytics_Settings $settings;
```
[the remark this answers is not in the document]
defunctl: No we don't, just in comments. Will fix this.

wp-content/plugins/core/functions/templates.php
```diff
@@ -10,7 +10,7 @@
 	 *
 	 * @return false|string
 	 */
-	function tribe_template_part( $slug, $name = null, $args = [] ) {
+	function tribe_template_part( string $slug, ?string $name = null, array $args = [] ) {
```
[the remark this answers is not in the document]
defunctl: This was just automated phpcbf based on the existing docblock, but I think this should actually enforce returning a string. Will update.

wp-content/plugins/core/src/Nav_Menus/Menu.php
```diff
@@ -6,49 +6,61 @@
...
-	private $cache_key = '';
+	/**
+	 * @var mixed[]
+	 */
+	private array $args;
+
+	/**
+	 * @var mixed[]
+	 */
+	private array $cache_key = [];
```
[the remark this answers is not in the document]
defunctl: Because it was never a string to begin with. I can double check this all works correctly though (or write some tests): https://github.com/moderntribe/square-one/blob/ba44b39ecabce2fc4c46be4b99187e55b8644c62/wp-content/plugins/core/src/Nav_Menus/Menu.php#L42-L45

wp-content/plugins/core/src/Query/Featured_Posts.php
```diff
@@ -12,28 +12,26 @@ class Featured_Posts {
 	 * this method, when used with the 'pre_get_post' action, will remove featured posts
 	 * from the main query to prevent posts from showing twice if we have featured posts on main blog page.
 	 *
-	 * @hook https://developer.wordpress.org/reference/hooks/pre_get_post/
+	 * @action pre_get_post
 	 */
-	public function remove_featured_posts_from_main_query( WP_Query $query ): WP_Query {
+	public function remove_featured_posts_from_main_query( WP_Query $query ): void {
```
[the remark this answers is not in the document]
defunctl: [pre_get_posts](https://developer.wordpress.org/reference/hooks/pre_get_posts/) is actually an action, this shouldn't have been returning anything in the first place.

wp-content/plugins/core/src/Routes/Routes_Subscriber.php
defunctl: It was duplicated notes from the method, so just providing extra noise: https://github.com/moderntribe/square-one/blob/ba44b39ecabce2fc4c46be4b99187e55b8644c62/wp-content/plugins/core/src/Routes/Routes_Definer.php#L14-L18

wp-content/plugins/core/src/Routes/Sample_REST_Route.php
defunctl: A comment that just repeats the class name that's right below it doesn't really provide any value and just adds noise to the code.

wp-content/plugins/core/src/Routes/Sample_Route.php
defunctl: Same deal, plus it's in a spot above the namespace which none of other code has ever done.

wp-content/plugins/core/src/Templates/Components/Abstract_Controller.php
defunctl: I will just type it with the controller :)

wp-content/plugins/core/src/Theme/Config/Image_Sizes.php
defunctl: I've only ever seen `string[]`, which would symbolize an array of strings, e.g. 

image

These are arrays with strings as keys, and mixed data as values (ints/bools).

I was being a bit lazy, so technically this should read `array<string, int|bool>` and now PhpStorm can warn you if you're using it for something other than what it's used for, because it now knows the exact type:

image

wp-content/plugins/core/src/Theme/Config/Web_Fonts.php
defunctl: According to [the existing comments I read](https://github.com/moderntribe/square-one/blob/9a65e37cc838f6a1d7c0a6180d98c0a0e2338644/wp-content/plugins/core/src/Theme/Theme_Definer.php#L119) these will be an array with a string as the key and a string as the value, so this is accurate.

These are PHPStan doc types: https://phpstan.org/writing-php-code/phpdoc-types

wp-content/plugins/core/src/Theme/Media/Image_Wrap.php
defunctl: Shouldn't be, I just added some defensive coding because technically `preg_replace_callback` COULD return an array, string or null, and this way if it fails, it'll just return the original content instead.

wp-content/plugins/core/src/Theme/Media/Oembed_Filter.php
defunctl: Automated, as mentioned way above :)

wp-content/plugins/core/src/Nav_Menus/Menu.php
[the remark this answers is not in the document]
defunctl: Confirmed this is working correctly.

image

wp-content/plugins/core/src/Theme/Branding/Login_Screen.php
nickpelton: #TIL https://developer.mozilla.org/en-US/docs/Web/API/HTMLStyleElement/type

nickpelton (approved):

PR #906, opened 2022-01-06
Run same validation as deploys do, make sure we now check pcss files

defunctl:
## What does this do/fix?

- The current eslint action does not run the same checks as a deploy does, so it often did not catch problems ahead of time. It was also not checking any pcss files.


defunctl:
Actually going to close this for now, some of the same work is being done in the CI job, maybe these just need to be combined.

PR #916, opened 2022-01-27
[SQONE-684] Swiper imagesReady event listener

LayaTaal:
## What does this do/fix?

In the slider component JS, the imagesReady event is added to the sliders after they have already been initialized. This leads to unpredictable triggering of the imagesReady event since in some cases, the slider images have already been loaded by the time the event is added.

This PR adds the event listener during initialization of a `Swiper` instance to ensure it is always triggered. It also updates the `modern_tribe/swiper_images_ready` custom event to pass the current swiper instance when that event is triggered, allowing easier access to the Swiper instance and all of its parameters and methods.

## Tests

Does this have tests?

- [ ] Yes


dpellenwood (approved):
:rocket: :dash:

PR #917, opened 2022-01-28
[SQONE-644] Fix social share pop for multiple components

LayaTaal:
## What does this do/fix?

The JS for the social-share component uses the first instance of itself as the container element. This prevents the JS events from being delegated to additional instances of the component (such as a single post with share icons in the header and footer). 

This PR allows for an indefinite number of social share components to be used on a page.

## QA

- [Link to Demo](https://sq1-pr.moderntribe.qa/post-49/)

### DIFF
--- a/wp-content/themes/core/components/share/js/share.js
+++ b/wp-content/themes/core/components/share/js/share.js
@@ -14,7 +14,7 @@ import * as tools from 'utils/tools';
 import popup from 'utils/dom/popup';
 
 const el = {
-	container: tools.getNodes( 'social-share-networks' )[ 0 ],
+	container: tools.getNodes( 'social-share-networks' ),
 };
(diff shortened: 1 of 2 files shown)


wp-content/themes/core/components/share/js/share.js
```diff
@@ -48,7 +48,7 @@ const bindEvents = () => {
  */
 
 const socialShare = () => {
-	if ( ! el.container ) {
+	if ( el.container.length === 0 ) {
```
LayaTaal, line 51: Note that since we are no longer setting `el.container` to the first item in the node array, `! el.container` evaluates as `true` even when no elements are present.

dpellenwood (approved):
:bug: :dash:

PR #918, opened 2022-01-28
Reduce unnecessary JavaScript initialization

LayaTaal:
## What does this do/fix?

There are a number of components that are able to initialize their JavaScript even when the component is not present on a page. This PR updates those components so that any required JS is only loaded when necessary.

**Components Addressed:**
- Accordion
- Card
- Tabs
- Gallery Slider Block
- Gravity Forms

## QA

- [Link to Demo](http://sq1-pr.moderntribe.qa/)

Screenshots/video:
- [Console Log showing init of non-present components](http://p.tri.be/M4vc9o)


dpellenwood (approved):
:elephant: :dash:

PR #925, opened 2022-01-31
Add proxy_ssl_server_name on; for nginx image proxy

defunctl:
## What does this do/fix?

If your production host for proxying images is behind Cloudflare, the image proxy will fail without this addition to support SNI SSL certificates.

## QA

- A production domain configured in nginx.conf that is behind Cloudflare will now display images on your local environment.

## Tests

Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests because it's an nginx update.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,7 +2,8 @@
 
 All notable changes to this project will be documented in this file.
 
-## 2022.01 
+## 2022.01
+* Added: `proxy_ssl_server_name on;` in nginx.conf to allow proxying to production domains if they are using Cloudflare.
 * Fixed: phpcs GitHub workflow using the wrong secret.
 * Added: An empty CLI_Definer to easier register commands.
 * Updated: added allowed plugins to composer.json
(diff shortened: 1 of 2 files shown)


dpellenwood:
Wonder if you might also want to add the host header value as I've had to do that for some SSL hosts (not Cloudflare specifically). For example, on Rakuten Rewards, this is my image proxy LOCATION RULESET:

defunctl:
> 

That's a good idea, I'll add it commented out with an explanation

defunctl:
Actually @dpellenwood, setting the same host header as the proxy pass should never be required, Nginx should be doing that out of the box, behind the scenes configuring and using the `$proxy_host` variable. 

At first I read your config thinking you needed a different Host header because of some weird DNS on Rakuten's side, but their DNS checks out. 

Your local DNS might be misconfigured because this is working for me without a Host header without issue:

image

nickpelton (approved):

dpellenwood:
:cucumber: :dash:

PR #932, opened 2022-02-02
Add `so project:test` custom command

defunctl:
## What does this do/fix?

Adds the custom command: `so project:test`, which runs all test suites so we don't have to run them all separately.

## Tests

Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests because it's an `so` custom command.


nickpelton (approved):
Tested locally and works great. 🚀

PR #945, opened 2022-03-04
update node version action to use node-version-file since previous wa…

nickpelton:
## What does this do/fix?

On a recently setup project an update to the node version setting on deploys was discovered to be broken. This should fix that.


dpellenwood (approved):
:black_cat: :dash:

PR #946, opened 2022-03-04
Fix cookie type issue

nickpelton:
## What does this do/fix?

In some cases `COOKIE_DOMAIN` will return a bool. This always casts it as a string JIC.


ChrisMKindred (approved):
🚢

dpellenwood (approved):
:shipit:

PR #951, opened 2022-03-09
[SQONE-546] Accordion Component A11y Improvements

clontztopher:
## What does this do/fix?

- Removes aria attributes associated with tab panels in favor of ones associated with accordion structure: [tabs](https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html) -> [accordion](https://www.w3.org/TR/wai-aria-practices-1.1/examples/accordion/accordion.html)
- Updates js to only set needed aria attributes for accordions

## QA

Links to relevant issues
- [SQONE-546](https://moderntribe.atlassian.net/browse/SQONE-546)

Screenshots/video:

## Tests

Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests because it was tested using Axe DevTools


dpellenwood (approved):
This is excellent, Chris. One additional request before merging: Can you remove the [accessibility.js utility module](https://github.com/moderntribe/square-one/blob/c9b37a619cb7457080a5a92e1a2fd0de3904f048/wp-content/themes/core/assets/js/src/utils/dom/accessibility.js)? We don't use those functions at all anywhere else in Sq1, so it's safe to delete that file.

clontztopher:
@dpellenwood Will do!

PR #952, opened 2022-03-09
[SQONE-514] Accordion Block/Component Refinements

clontztopher:
## What does this do/fix?

- Adds ability to enable/disable scroll-to functionality for accordion block

## QA

Links to relevant issues
- [SQONE-514](https://moderntribe.atlassian.net/browse/SQONE-514)

## Tests

Does this have tests?

- [ ] Yes

### DIFF
--- a/wp-content/themes/core/components/accordion/js/accordion.js
+++ b/wp-content/themes/core/components/accordion/js/accordion.js
@@ -65,12 +65,24 @@ const setOffset = () => {
 };
 
 /**
+ *
+ * @function triggerAfterScroll
+ * @description - Trigger the `modern_tribe/accordion_animated` event
+ */
+const triggerAfterScroll = ( data ) => () => events.trigger( {
+	event: 'modern_tribe/accordion_animated',
+	native: false,
+	data,
+} );
+
+/**
  * @function openAccordion
  * @description Toggle the accordion open
  */
 
 const openAccordion = ( header, content ) => {
 	const row = tools.closest( header, '.c-accordion__row' );
+	const accordion = tools.closest( header, '.c-accordion' );
 	closeOthers( row );
 	tools.addClass( row, 'active' );
 	header.setAttribute( 'aria-expanded', 'true' );
@@ -81,25 +93,16 @@ const openAccordion = ( header, content ) => {
 	slide.down( content, content.id, options.speed );
 
 	_.delay( () => {
-		if ( ! el.container[ 0 ].dataset.scrollto ) {
-			events.trigger( {
-				event: 'modern_tribe/accordion_animated',
-				native: false,
-			} );
-		} else {
-			scrollTo( {
-				after_scroll: () => {
-					events.trigger( {
-						event: 'modern_tribe/accordion_animated',
-						native: false,
-					} );
-				},
-
-				offset: options.offset,
-				duration: 300,
-				$target: $( row ),
-			} );
+		if ( ! accordion.dataset.scrollto ) {
+			triggerAfterScroll();
+			return;
 		}
+		scrollTo( {
+			afterScroll: triggerAfterScroll(),
+			offset: options.offset,
+			duration: 300,
+			$target: $( row ),
+		} );
 	}, options.speed );
 };
 
@@ -144,7 +147,6 @@ const handlePanelEvents = ( e ) => {
  * @param {object} e The js event object.
  * @description Toggle the active accordion item using class methods.
  */
-
 const toggleItem = ( e ) => {
 	const header = e.delegateTarget;
 	const row = tools.closest( header, '.c-accordion__row' );
--- a/wp-content/themes/core/components/accordion/js/accordion.js
+++ b/wp-content/themes/core/components/accordion/js/accordion.js
@@ -80,23 +80,26 @@ const openAccordion = ( header, content ) => {
 
 	slide.down( content, content.id, options.speed );
 
-	if ( ! el.container[ 0 ].dataset.scrollto ) {
-		return;
-	}
-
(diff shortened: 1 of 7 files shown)


wp-content/themes/core/components/blocks/accordion/Accordion_Block_Controller.php
```diff
@@ -51,6 +52,8 @@ class Accordion_Block_Controller extends Abstract_Controller {
 	 */
 	private array $cta;
 
+	private bool $scroll_to;
```
dpellenwood, line 55:
```suggestion
	/**
	 * @var bool
	 */
	private bool $scroll_to;
```

wp-content/themes/core/components/accordion/js/accordion.js
```diff
@@ -79,6 +79,11 @@ const openAccordion = ( header, content ) => {
 	setOffset();
 
 	slide.down( content, content.id, options.speed );
+
+	if ( ! el.container[ 0 ].dataset.scrollto ) {
```
dpellenwood: Unfortunately, this selector won't work in instances where more than one accordion component is used on the page. `el.container[ 0 ]` will always be the first, regardless of which one is clicked.

You'll probably want to set a new const for the _current_ component using `closest()` and then check that for the data attr.

wp-content/themes/core/components/blocks/accordion/Accordion_Block_Controller.php
```diff
@@ -25,6 +25,7 @@ class Accordion_Block_Controller extends Abstract_Controller {
 	public const LEADIN            = 'leadin';
 	public const ROWS              = 'rows';
 	public const TITLE             = 'title';
+	public const SCROLL_TO 		   = 'scroll_to';
```
dpellenwood: Check your alignment here. :)

wp-content/themes/core/components/blocks/accordion/Accordion_Block_Controller.php
```diff
@@ -114,7 +118,10 @@ public function get_header_args(): array {
 
 	public function get_content_args(): array {
 		return [
-			Accordion_Controller::ROWS => $this->rows,
+			Accordion_Controller::ROWS            => $this->rows,
+			Accordion_Controller::CONTAINER_ATTRS => [
+				'data-scrollto' => $this->scroll_to,
+			],
```
dpellenwood, line 124: Right now, the necessary html attribute is added via the block, but this feature should also work at the component level (in cases where we use an accordion component outside the block context).

To make this work at the component level, you'll need to add a similar const, property, & attribute to the component controller. Then, rather than output the HTML attribute here, you'll want to do that inside the accordion component as one of it's conditional attributes instead of the block's.

wp-content/themes/core/components/accordion/js/accordion.js
```diff
@@ -79,19 +79,27 @@ const openAccordion = ( header, content ) => {
...
+						event: 'modern_tribe/accordion_animated',
+						native: false,
+					} );
+				},
+
+				offset: options.offset,
+				duration: 300,
+				$target: $( row ),
+			} );
+		}
```
dpellenwood, line 102: A couple things here to note:

1. `el.container[ 0 ]` will always be the first instance of an accordion component on the page, even when there are multiple instances. You'll have to adjust this to "find" the specific accordion container for the `header` that was clicked rather than just assuming the `0` instance.
2. We try to avoid else conditions as often as possible because it requires extra thought processing to grep and more room for errors. I would suggest an alternative here would be to do a `return;` after the even trigger in the first condition. Then the scroll to (else condition) can just exist as stand alone statements without the else condition wrapping it.
3. it's always good to reduce code duplication into separate methods whenever possible. I'd recommend moving the event trigger to it's own method and calling it in each conditional context rather than repeating it. (Notice I also suggest updated the `closeAccordion()` method since it too triggers the same event.)

Here's an example of these changes. Be sure to test them though. I didn't run them them anywhere before creating it:
```

/**
 * @function triggerAfterScroll
 * @description Handles triggering the accordion animated event.
 */
const triggerAfterScroll = () => {
	events.trigger( {
		event: 'modern_tribe/accordion_animated',
		native: false,
	} );
};

/**
 * @function openAccordion
 * @description Toggle the accordion open
 */

const openAccordion = ( header, content ) => {
	const component = tools.closest( header, '[data-js="c-accordion"]' );
	const row = tools.closest( header, '.c-accordion__row' );
	closeOthers( row );
	tools.addClass( row, 'active' );
	setAccActiveAttributes( header, content );
	content.removeAttribute( 'hidden' );
	setOffset();

	slide.down( content, content.id, options.speed );

	if ( ! component.dataset.scrollto ) {
		triggerAfterScroll();
		return;
	}

	_.delay( () => {
		scrollTo( {
			after_scroll: () => triggerAfterScroll,
			offset: options.offset,
			duration: 300,
			$target: $( row ),
		} );
	}, options.speed );
};

/**
 * @function closeAccordion
 * @description Toggle the accordion closed
 */

const closeAccordion = ( header, content ) => {
	const row = tools.closest( header, '.c-accordion__row' );
	tools.removeClass( row, 'active' );
	setAccInactiveAttributes( header, content );
	slide.up( content, content.id, options.speed );
	_.delay( () => {
		content.setAttribute( 'hidden', 'true' );
		triggerAfterScroll();
	}, options.speed );
};
```

dpellenwood (changes requested):
Good work, Chris. I have a few requested changes, but you've got the right idea here. Hit me up if you have any questions. Thank you!

dpellenwood (approved):
:shipit:

PR #957, opened 2022-03-18
Feature: remove blocks by page template

defunctl:
## What does this do/fix?

- Allows developers to specify blocks that will only be selectable when certain WordPress page templates have been selected.

## QA

Using this as a test example in the definer:

```php
			Block_Page_Template_Filter::class => DI\autowire()
				->constructor( [
					 'page-templates/page-example-app.php' => [
						 Content_Loop::NAME,
					 ],
				] ),
```

Default template, Content Loop is not available:

image

Change to the Example React App Template and the block becomes available in real time:

image

**Note:** Once the block is on there, this doesn't handle removing it or antyhing.

## Tests

Does this have tests?

- [x] Yes

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -3,6 +3,7 @@
 All notable changes to this project will be documented in this file.
 
 ## 2022.03
+* Added: Block Page Template Filter: Allow certain blocks to only be chosen when using a specific page template.
 * Changed: Replaced `msawicki/acf-menu-chooser` with a forked https://github.com/moderntribe/acf-menu-chooser that includes security fixes and is also added to packagist.
 * Updated: ACF (5.12), Tribe Libs (3.4.10), Redirection (5.2.3), Yoast (18.2), TEC (5.14.0.4)
 * Updated: Aligns accordion component with WAI-ARIA standard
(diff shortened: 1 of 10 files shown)


ChrisMKindred (commented):
This looks good and will be great to have.  I just have one question and maybe I am missing it in the JS... 
What would happen if you were to enter a block then change the template when that original block was one listed in the `block_page_template_filter` for the new template selected?

defunctl:
> This looks good and will be great to have. I just have one question and maybe I am missing it in the JS... What would happen if you were to enter a block then change the template when that original block was one listed in the `block_page_template_filter` for the new template selected?

Thanks for pointing this out, I had a note in the PR that it didn't do anything once the block is there, but it actually crashes Gutenberg if you switch templates if a filtered block is on the page.

Any suggestions on how we should best handle this for a user experience perspective? I think finding a way to warn them that they'll lose their blocks or something like that if they try to switch templates after they have filtered blocks on the page might be decent, allowing them to cancel that action.

ChrisMKindred:
> Any suggestions on how we should best handle this for a user experience perspective? I think finding a way to warn them that they'll lose their blocks or something like that if they try to switch templates after they have filtered blocks on the page might be decent, allowing them to cancel that action.

I think a warning message with a cancel/continue option would be great.  Then remove the block from the content if they continue.

defunctl:
We are going to update this in our project first, @ScottGarrison is going to have a look :)

dpellenwood:
Closing as SquareOne is no longer under active development.

PR #961, opened 2022-03-24
Bugfix/gravityforms filter param types

dpellenwood:
## What does this do/fix?
Gravity Forms v2.5 and above have changed the parameter types being passed through some of their filters. This PR updates our code to use the correct types. 

- [ ] Yes
- [x] No, this doesn't need tests because it's an integration fix.


nickpelton (approved):

wp-content/plugins/core/src/Integrations/Gravity_Forms/Form_Markup.php
```diff
@@ -16,14 +16,14 @@ class Form_Markup {
...
-	 * @param string $value
+	 * @param string       $choice_markup
+	 * @param array        $choice
+	 * @param object       $field
+	 * @param array|string $value // GF docs says this should be a string, but both strings and arrays may be returned as of v2.5
 	 *
 	 * @return string
 	 */
-	public function customize_gf_choice_other( string $choice_markup, array $choice, array $field, string $value ): string {
+	public function customize_gf_choice_other( string $choice_markup, array $choice, object $field, $value ): string {
```
defunctl, line 26: Looks like we're still using this as an array below, why don't we remove the type (and update the param to be `array|object` in the DocBlock) and then just force cast it as an array?

something like: 

```php
/**
 * @param object|array       $field
 * /
public function customize_gf_choice_other( string $choice_markup, array $choice, $field, $value ): string {
    $field = (array) $field;
    // ...more code
}
```

defunctl (changes requested):

wp-content/plugins/core/src/Integrations/Gravity_Forms/Form_Markup.php
```diff
@@ -51,12 +52,12 @@ public function customize_gf_choice_other( string $choice_markup, array $choice,
...
 	 *
 	 * @param string $classes
-	 * @param array  $field
+	 * @param object $field
 	 * @param array  $form
 	 *
 	 * @return string
 	 */
-	public function add_gf_select_field_class( string $classes, array $field, array $form ): string {
+	public function add_gf_select_field_class( string $classes, object $field, array $form ): string {
```
defunctl, line 60: @dpellenwood don't forget this one.

dpellenwood: Thanks @defunctl , Fixed. Can you give this another quick look-see?

defunctl (approved):
:truck: :dash:

PR #964, opened 2022-03-29
[BUGFIX]: Content loop pagination and Index controller refactor

defunctl:
## What does this do/fix?

- The archive/post index loops currently have no pagination and no way to view more posts. This adds pagination to the content loop block component (with an option to disable it, like for featured posts).
- Lots of other bug fixes / code clean up. See the changelog updates.

## QA

Homepage with featured posts, 5 posts per page and pagination:
image

Tag Archive with custom image:
image

Category archive with custom image:
image

Page two of category archive:
image

## Tests

Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests because it's controller/component updates.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -36,6 +36,7 @@ tab_width = 4
 
 [wp-content/{plugins,themes}/core/**.php]
 indent_style = tab
+indent_size = 4
 ij_formatter_off_tag = @formatter:off
 ij_formatter_on_tag = @formatter:on
 ij_formatter_tags_enabled = false
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,7 @@ local-config.php
 cache-config.php
 tests-config.php
 build-process.php
+wp-content/db.php
 wp-content/uploads/
 wp-content/blogs.dir/
 wp-content/upgrade/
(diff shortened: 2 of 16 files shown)


nickpelton (approved):
Nice cleanup! 🚀

dpellenwood (commented):
@defunctl , I love this. However, I'm not a fan of the content_loop block being used in the route  templates (index, archive, etc) at all. It's unnecessarily nested. I've had a ticket in the backlog to unwind that architecture complete from the loop routes.

nickpelton:
> @defunctl , I love this. However, I'm not a fan of the content_loop block being used in the route templates (index, archive, etc) at all. It's unnecessarily nested. I've had a ticket in the backlog to unwind that architecture complete from the loop routes.
> 
> Thoughts on just removing it completely and allowing the routes to handle their loops and pagination in a more traditional sense?

IIRC this was like this before these improvements. So should we tackle this as part of that other ticket you mentioned?

dpellenwood (approved):
> I'm down for a clean up, but if that's not happening any time soon, the framework as it is right now is pretty broken without the fixes I just put there and any new projects will basically have broken loops and no way to see more posts than the initial page load

PR #965, opened 2022-03-30
Dokku & Docker config updates

dpellenwood:
## What does this do/fix?
* Aligns Dokku NGINX client_max_body_size and PHP max_filesize, and post_max_size
* Adds support for webp and woff2 static file handling to both Docker and Dokku
* Adds a few PHP extensions that WordPress recommends/requires so that Dokku environments are setup properly for WordPress.

### DIFF
--- a/dev/docker/nginx/nginx.conf
+++ b/dev/docker/nginx/nginx.conf
@@ -56,7 +56,7 @@ server {
 
         try_files $uri $uri/ @rewrites;
 
-        location ~* \.(?:ico|css|js|gif|jpe?g|png|svg|eot|ttf|woff|mp4|m4v|mov|wmv|avi|mpg|ogv|3gp|3g2|flv|webm|aac|m4a|f4a|mp3|ogg|oga)$ {
+        location ~* \.(?:ico|css|js|gif|jpe?g|png|svg|eot|ttf|woff2?|mp4|m4v|mov|wmv|avi|mpg|ogv|3gp|3g2|flv|webm|aac|m4a|f4a|mp3|ogg|oga|webp)$ {
             try_files $uri =404;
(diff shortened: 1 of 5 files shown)


dev/docker/nginx/nginx.conf
```diff
@@ -56,7 +56,7 @@ server {
 
         try_files $uri $uri/ @rewrites;
 
-        location ~* \.(?:ico|css|js|gif|jpe?g|png|svg|eot|ttf|woff|mp4|m4v|mov|wmv|avi|mpg|ogv|3gp|3g2|flv|webm|aac|m4a|f4a|mp3|ogg|oga)$ {
+        location ~* \.(?:ico|css|js|gif|jpe?g|png|svg|eot|ttf|woff2?|mp4|m4v|mov|wmv|avi|mpg|ogv|3gp|3g2|flv|webm|aac|m4a|f4a|mp3|ogg|oga|webp)$ {
```
defunctl, line 59: @dpellenwood  If we're proxying this stuff from a remote server, wouldn't we want both: `woff` and `woff2`?

dpellenwood: Yep. the question mark makes the `2` optional for that matching pattern. So both `.woff` and `.woff2` would match.

defunctl: hrmm wouldn't that be `woff?2`, can't remember the details of nginx regexes.

defunctl (approved):

dev/docker/nginx/nginx.conf
dpellenwood: Thanks! I double-checked. Nginx use the PCRE format for regexes. ;)

PR #966, opened 2022-03-30
update to latest docker php image with some new packages. …

nickpelton:
## What does this do/fix?

Update to latest PHP docker images that have some new php extensions installed

https://github.com/moderntribe/squareone-docker-images/pull/14 
https://github.com/moderntribe/squareone-docker-images/pull/16


nickpelton:
#TIL Since SQ1 is pointing at moderntribe/squareone-php:74-3.0, if you pull it will auto grab a patch on a docker pull

PR #967, opened 2022-03-30
Add missing pr number to review app workflow

nickpelton:
## What does this do/fix?

* Missing Review app PR number after we abstracted the GHA tasks to `moderntribe/actions`
* Also removed any push deploy events, those should be opt-in when the project is setup.


marcossantiago (approved):
I have fixed passing the PR_NUMBER to another action and tested. It should be fine now. 

This will fail though, because it's pointin to the `main` branch action which has been updated but will be fixed once merged. 

🚢

PR #973, opened 2022-04-08
disable dependabot

nickpelton:
## What does this do/fix?

Dependabot has been noisy and the dependency management is cluttered with PRS per change which makes it hard to test. Also, individual projects having dependabot on by default is unhelpful. I considered this a failed experiment in managing our dependencies better, so I'm suggesting we remove this for now.


ChrisMKindred (commented):
Agreed.

ChrisMKindred (approved):

PR #976, opened 2022-04-20
Webpack & related updates for React app dev (Example et al.)

dpellenwood:
## What does this do/fix?
Updates some of the underlying webpack and related config to get HMR dev working properly with our local dev certs and hostnames. Also some changes due to updated APIs in webpack.

### DIFF
--- a/wp-content/plugins/core/src/Assets/Theme/Scripts.php
+++ b/wp-content/plugins/core/src/Assets/Theme/Scripts.php
@@ -119,7 +119,7 @@ class Scripts {
 			$this->localize_scripts( (string) reset( $handles ) );
 
 			if ( defined( 'HMR_DEV' ) && HMR_DEV === true ) {
-				wp_enqueue_script( 'tribe-scripts-hmr-bundle', 'http://localhost:3000/app.js', $handles, time(), true );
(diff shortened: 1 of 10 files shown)


wp-content/plugins/core/src/Assets/Theme/Scripts.php
```diff
@@ -119,7 +119,7 @@ public function enqueue_scripts(): void {
 			$this->localize_scripts( (string) reset( $handles ) );
 
 			if ( defined( 'HMR_DEV' ) && HMR_DEV === true ) {
-				wp_enqueue_script( 'tribe-scripts-hmr-bundle', 'http://localhost:3000/app.js', $handles, time(), true );
+				wp_enqueue_script( 'tribe-scripts-hmr-bundle', home_url() . ':9000/app.js', $handles, time(), true );
```
jorgemasta, line 122: where does `home_url` come from?

jorgemasta (approved):

wp-content/plugins/core/src/Assets/Theme/Scripts.php
nickpelton: WordPress Options table homeurl value. Or the environment variable `define('WP_HOME')`

nickpelton (approved):

PR #977, opened 2022-04-21
Feature/Discussion: ACF Maxlength Helper Function

ChrisMKindred:
## What does this do/fix?

Adds an ACF helper class that triggers a max character counter on input or textarea fields that include a maxlength value. For every project I have worked on, there has been a request for this feature and the level of discovery/lift always kept us from getting it added. 

I have included example usage in the video below.

## QA

Screenshots/video:
- [Link to Video](http://p.tri.be/4nsXJG)

## Tests

Does this have tests?

- [x] Yes


dpellenwood (approved):
Looks good. Nice addition!

PR #979, opened 2022-04-27
Fix Gravity Forms undefined index errors

jamiepaul:
## What does this do/fix?

Many undefined index errors occur in the Gravity Forms form editor when adding fields to a form. `inputType` is no longer passed w/ field data and field type enhancements (i.e. enableEnhancedUI for select dropdowns) only have the corresponding key when checked. 

## Tests

Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests because fixes an integration

### DIFF
--- a/wp-content/plugins/core/src/Integrations/Gravity_Forms/Form_Markup.php
+++ b/wp-content/plugins/core/src/Integrations/Gravity_Forms/Form_Markup.php
@@ -68,7 +68,7 @@ class Form_Markup {
 		} elseif ( $field['type'] === 'select' ) {
 			$classes .= ' gf-select';
 			// Not Chosen, regular select
-			if ( array_key_exists( 'enableEnhancedUI', $field ) && ! $field['enableEnhancedUI'] ) {
+			if ( empty( $field['enableEnhancedUI'] ) ) {
 				$classes .= ' gf-select-no-chosen';
 			}
 		} elseif ( $field['type'] === 'checkbox' ) {
--- a/wp-content/plugins/core/src/Integrations/Gravity_Forms/Form_Markup.php
+++ b/wp-content/plugins/core/src/Integrations/Gravity_Forms/Form_Markup.php
@@ -63,38 +63,38 @@ class Form_Markup {
 		$class_icon_simple  = $this->activate_icons ? ' form-control-icon' : '';
 		$class_icon_complex = $this->activate_icons ? ' form-control-icon-complex' : '';
 
-		if ( $field['type'] === 'multiselect' || $field['inputType'] === 'multiselect' ) {
+		if ( $field['type'] === 'multiselect' ) {
 			$classes .= ' gf-multi-select';
-		} elseif ( $field['type'] === 'select' || $field['inputType'] === 'select' ) {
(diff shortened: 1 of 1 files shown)


dpellenwood (commented):
Interesting @jamiepaul, I haven't run into this before. Which version of WordPress and which version of Gravity Forms are you testing?

jamiepaul:
> Interesting @jamiepaul, I haven't run into this before. Which version of WordPress and which version of Gravity Forms are you testing?

@dpellenwood I saw it pop up on our Cleveland Community College project, which is pretty new - WP v5.9.3 and GF v2.6.1.1 and was able to reproduce it on my local for SQ1

To create the issue, I just create a new form and on the editor try and drag a text field to it which results in the error. Here's a [screencast](http://p.tri.be/1KGrAy) from the `main` branch of SQ1, WP v5.9.2 and GF v2.6.1.1

dpellenwood (approved):
Gotcha. Thanks for debugging and porting back to Sq1, @jamiepaul!

defunctl (approved):
Looks like this should work. 

Do we have to worry about any backwards compatibility at all here? I'm guessing not as new projects would clone this and also get the latest Gravity Forms at the same time, so they'd be up to date together.

I think this whole thing could be made into an array to be much much cleaner but that's for another time :grin:

wp-content/plugins/core/src/Integrations/Gravity_Forms/Form_Markup.php
```diff
@@ -63,38 +63,38 @@ public function add_gf_select_field_class( string $classes, $field, array $form
...
 
-		if ( $field['type'] === 'multiselect' || $field['inputType'] === 'multiselect' ) {
+		if ( $field['type'] === 'multiselect' ) {
 			$classes .= ' gf-multi-select';
-		} elseif ( $field['type'] === 'select' || $field['inputType'] === 'select' ) {
+		} elseif ( $field['type'] === 'select' ) {
 			$classes .= ' gf-select';
 			// Not Chosen, regular select
-			if ( ! $field['enableEnhancedUI'] ) {
+			if ( array_key_exists( 'enableEnhancedUI', $field ) && ! $field['enableEnhancedUI'] ) {
```
defunctl, line 71:
```suggestion
			if ( empty( $field['enableEnhancedUI'] ) ) {
```
This one could be simplified since we're just checking for a negative value

jamiepaul:
@dpellenwood @defunctl thank you both!

For backwards compatibility, I also assumed it was a non-issue as new projects would clone and also get the latest GF along with this change.

Any tips on getting the last sniff to run so I can merge? It's said "Expected — Waiting for status to be reported" for about 10mins now

ChrisMKindred:
> Any tips on getting the last sniff to run so I can merge? It's said "Expected — Waiting for status to be reported" for about 10mins now

It looks like it [did pass](https://github.com/moderntribe/square-one/runs/6202015604?check_suite_focus=true), but for some reason isn't passing the value back in for the checks. All PRs are doing it right now.  Is it not letting you merge without it passing?

nickpelton:
Looks like a GHA bug. I merged.

PR #981, opened 2022-05-06
Fix/apply loading attr conditionally

dpellenwood:
## What does this do/fix?
WP Core applies the `loading="lazy"` attribute to all `` tags (except the first) in `the_content` including our Image Component images. This fix removes that attribute for any instance of the image component `c-image__image`. This works b/c if we _do_ want lazy loading, we apply it via our 3rd-party lib, not the native implementation.

**Note:** In the future, I'd like to deep-dive on our component and reduce it or allow WP Core to handle image tags entirely, but for now, this is just a fix for our super-component.

Additionally, this update removes some old, unused image tag filtering and related unit tests from the TinyMCE days.


ChrisMKindred (approved):
Looks Good. I like that we are removing some of the tinymce filterings.

wp-content/plugins/core/src/Theme/Media/Image_Attributes.php
```diff
@@ -0,0 +1,31 @@
...
+	 *                             the attribute being omitted for the image.
+	 * @param string      $image   The HTML `img` tag to be filtered.
+	 * @param string      $context Additional context about how the function was called or where the img tag is.
+	 *
+	 * @filter wp_img_tag_add_loading_attr
+	 *
+	 * @return string|bool
+	 */
+	public function customize_wp_image_loading_attr( $value, string $image, string $context ) {
+		if ( 'the_content' === $context && strpos( $image, 'c-image__image' ) ) {
```
defunctl, line 24: Extremely unlikely, but this condition would return true if I uploaded an image called `c-image__image-whatever.png`. :smile:

defunctl (approved):
:+1: :taco: :dash:

wp-content/plugins/core/src/Theme/Media/Image_Attributes.php
dpellenwood: True. I think the only way we could avoid this would be to add a custom data attribute to every image component render. Considering that the worst-case outcome of the edge case is that the loading attribute is skipped, I'm okay with the risk.

PR #982, opened 2022-05-06
Make sure lazyloaded images are loaded before printing.

dpellenwood:
## What does this do/fix?
Lazy loaded images were not being rendered for print. This bit of script makes sure the images get loaded before the print dialog is triggered.


ScottGarrison (approved):

wp-content/themes/core/components/image/js/image.js
```diff
@@ -0,0 +1,33 @@
...
+};
+
+/**
+ * @function init
+ */
+const init = () => {
+	bindEvents();
+
+	console.info( 'SquareOne Theme: Initialized Image component.' );
+};
```
ScottGarrison, line 31: Should we check to make sure there's lazy load images on the page before kicking off the script?

dpellenwood: Good question. I considered that and opted against it for 2 reasons:

jamiepaul (approved):
🖨️ 💨

Vinsanity (commented):
LGTM! 🎠 🍬

PR #984, opened 2022-05-12
Feature: quote + image block field model improvements

defunctl:
## What does this do/fix?

- As per the back-end meeting, no reason not to refactor the Quote + Image Block as demo'd. Moved all the data into its own group to simplify the field model.
- Bugfix for accidentally applying the light theme when a hero image **didn't** exist, when it should be opposite.

## QA

image
image

## Tests

Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests because it's a block update.


lindseyolson (approved):

ChrisMKindred (approved):
👍

PR #985, opened 2022-05-17
Feature/autowired meta definer

defunctl:
## What does this do/fix?

- Use autowiring for object meta definers so we can use automatic dependency injection in the constructor.

## QA

- Should work exactly as before, but now you can auto inject instances into object meta classes.

## Tests

Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests because it's just a DI definition update.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -3,6 +3,7 @@
 All notable changes to this project will be documented in this file.
 
 ## 2022.05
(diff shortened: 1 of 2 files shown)


ChrisMKindred (approved):
👍

PR #990, opened 2022-05-25
Set up the container for use in integration tests

defunctl:
## What does this do/fix?

- Allows the use of `$this->container->get( Whatever::class )` for integration tests. Sometimes you want to actually test the class as the container builds it instead of creating the dependencies manually.

## QA

- Should function exactly as before.

## Tests

Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests because it's updating the test framework itself!

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -3,6 +3,7 @@
 All notable changes to this project will be documented in this file.
 
 ## 2022.05
+* Added: Access to the container in integration tests via `$this->container`.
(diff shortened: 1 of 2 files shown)


ChrisMKindred (approved):

dpellenwood (commented):
Looks good. I'll leave it as just a comment as this is more Chris' wheelhouse. :)

PR #992, opened 2022-05-25
Allow unlimited memory when running PHPStan via CI

defunctl:
## What does this do/fix?

- Allows PHPStan to use unlimited memory, at a certain number of files it will always run out and throw errors.

## QA

- No more failing CI.

## Tests

Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests because it's a CI update.


ChrisMKindred (approved):
👍

dpellenwood (approved):
Nice!

PR #995, opened 2022-06-10
Bugfix: Tab Block Controller type errors

defunctl:
## What does this do/fix?

- I missed converting the Tabs Block Controller to use the CTA field model and Tab Collection, which was causing type errors and making the block not usable.

## QA

Working Tabs Block Edit Screen
image

Working Tab Block Frontend
image

## Tests

Does this have tests?

- [ ] Yes
- [ ] No, this doesn't need tests because we don't have tests for blocks.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -3,6 +3,7 @@
 All notable changes to this project will be documented in this file.
 
 ## 2022.06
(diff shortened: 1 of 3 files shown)


ChrisMKindred (approved):
Thanks!

PR #997, opened 2022-06-14
[SQONE-574]: clear floats on all blocks

GeoffDusome:
## What does this do/fix?

https://moderntribe.atlassian.net/browse/SQONE-574

Blocks on both the front end & back end weren't properly being cleared around floated elements. Since all of our blocks are made to be full width, we can safely add `clear: both;` to all of our blocks.

(sorry about the commit having the wrong ticket tag, that's my bad)

## QA

1. Add a floated element
2. Add a SO block below the floated element, this block shouldn't float up beside the floated element.

## Tests

Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests because it's styling.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -3,6 +3,7 @@
 All notable changes to this project will be documented in this file.
 
 ## 2022.06
+* Fixed: Blocks added below floated elements in Gutenberg should now properly clear on both the frontend and backend.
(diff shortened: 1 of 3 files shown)


dpellenwood (approved):
🪲 💨

PR #999, opened 2022-06-14
Chore/wp core plugin updates [DO NOT MERGE]

dpellenwood:
## What does this do/fix?
* Updates WordPress core to v6.0
* Updates misc plugins

HSPH has been running 6.0 during development and didn't come across any issues. Probably good to make sure any new projects use the latest major version.


changelog.md
```diff
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 
 ## 2022.06
 * Fixed: `update-query-var.js` can now properly remove keys with `undefined` values.
+* Updated: WordPress to 6.0, misc plugin updates.
```
defunctl, line 7: Newer entries should be at the top so it reads newest updates to oldest :)

lindseyolson (approved):

defunctl (approved):
Approving, up to you if you want to make the changelog suggestion but I find it much more logical to scan based on top to bottom.

Vinsanity (approved):
Should we be waiting for 6.1 to do this?

dpellenwood:
> Should we be waiting for 6.1 to do this?
@Vinsanity , We've been using 6.0 on a variety of pre-launch / in-development projects for a while now. It's safe for new projects. Our general rule of thumb is that *new* projects should start (or update to) the latest version while in development. This allows them to stay up-to-date with Core versions as they progress and launch with the most-recent version of Core available. 

For already launched projects (those in maintenance), yes, we would prefer to wait for the first bug fix release after a major version, 6.0.1 in this case.

The difference is that already live projects have more risk for bugs or regressions not being found during a standard QA/smoketest cycle, and pre-launch projects have more space/time to address any issues since the site isn't live yet.

dpellenwood:
> if a theme is using a theme.json file there are markup changes to the image & video WP blocks which breaks our alignment styles for media in SQ1.

Thanks @jamiepaul ! Good to know! I'll be sure to address that here before merging.

ChrisMKindred (approved):
👍

dpellenwood:
Closing this as we'll want to wait to update core to 6.0.1 or 6.1 when [the underlying issue](https://core.trac.wordpress.org/ticket/54905) is addressed. And, we'll want to bump plugins before then.

PR #1001, opened 2022-06-17
RFC: Portable Blocks 'Step 1'

ChrisMKindred:
## What does this do/fix?

SquareOne folder structure for blocks is too convoluted. One of our goals over the next year is to provide a better way for our blocks to be portable between projects. We can make it much easier by narrowing the number of folders we have to grab when adding a block to a project. As the first step in this process, we would like to remove the block template routing folder from the theme. 

This RFC suggests that we start using the `render_template` key provided by ACF, and remove the Theme Blocks folder. This provides a cleaner structure by including the Route in the same folder as the model. Once we convert all blocks, we could remove the `Tribe_Libs/ACF/block_renderer`. We would also update the block generator to include the `[Block Name]_Route.php`.

Step 2: 
Move `wp_themes/core/components/blocks` to `wp_themes/core/blocks` evaluating our build process to match.


ChrisMKindred:
closed in favor of #1008

PR #1006, opened 2022-06-23
[IN_PROGRESS] Replaces translations with escape and translate in blocks

alexander-kozhukhovski:
## What does this do/fix?

- Replaces plain translations with escaping + translation for blocks

## QA

N/A

## Tests

Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests


Vinsanity (approved):
🥜 🍫

wp-content/plugins/core/src/Blocks/Types/Gallery_Slider/Gallery_Slider.php
```diff
@@ -51,30 +51,30 @@ public function add_fields(): void {
...
 					'type'  => 'link',
 				] )
 			)->add_field( new Field( self::NAME . '_' . self::GALLERY, [
-					'label'        => __( 'Gallery', 'tribe' ),
+					'label'        => esc_html__( 'Gallery', 'tribe' ),
 					'name'         => self::GALLERY,
 					'type'         => 'gallery',
 					'max'          => 12,
-					'instructions' => __(
+					'instructions' => esc_html__(
```
ChrisMKindred, line 77: 🔴 
The text here includes HTML. In this case the <br> will be output as `&lt;br&gt;`. We might need to break this down into multiple esc_html__() functions to allow for new lines in the correct spaces.

ChrisMKindred (changes requested):
It looks like Logo has some extra HTML in it for testing and we need to figure out a different solution for another description that includes HTML.  Everything else looks great.

wp-content/plugins/core/src/Blocks/Types/Gallery_Slider/Gallery_Slider.php
alexander-kozhukhovski: Thank you for noticing 👍

defunctl (approved):
Nice, thanks for doing this!

defunctl:
Companion PR for tribe-libs here: https://github.com/moderntribe/tribe-libs/pull/121

ChrisMKindred (approved):
Looks good!

PR #1007, opened 2022-06-23
Feature: unregister block styles

alexander-kozhukhovski:
## What does this do/fix?

- Removes PHP based functionality for unregistering block styles
- Adds functionality for specifying blocks and their styles for unregistering
- Provides WP filter for it
- Unregisters specified blocks styles from JS

To remove specific core block styles from editor add the records to the `\Tribe\Project\Blocks\Blocks_Definer::DENY_BLOCK_STYLES` like this:
```php
self::DENY_BLOCK_STYLES => [
    'core/separator' => [ 'default', 'wide', 'line', 'dots' ],
]
```

## QA

N/A

## Tests

Does this have tests?

- [ ] Yes

### DIFF
--- a/wp-content/plugins/core/src/Blocks/Block_Style_Deny_List.php
+++ b/wp-content/plugins/core/src/Blocks/Block_Style_Deny_List.php
@@ -1,0 +1,37 @@
+<?php declare(strict_types=1);
+
+namespace Tribe\Project\Blocks;
+
+/**
+ * Class Block_Style_Deny_List
+ *
+ * Provides an array of block names together with styles to remove from the editor.
+ */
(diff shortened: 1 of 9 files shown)


wp-content/plugins/core/src/Blocks/Block_Style_Deny_List.php
```diff
@@ -0,0 +1,37 @@
+<?php declare(strict_types=1);
```
ChrisMKindred, line 1: 🌮 I love how small and simple this class is!

ChrisMKindred (approved):
Thanks!

defunctl (approved):
:+1:

PR #1009, opened 2022-06-29
[SQONE-609] Update Sink Styles Specificity & Add Support for Columns Block

LayaTaal:
## What does this do/fix?
This PR adds support for the core columns block and in the process refactors base sink styles in many places to improve specificity.

### Updated Sink Styles Specificity
The main changes here are to how we handle applying sink styles to typography. The idea here is to reduce the specificity as much as possible to allow for easily overriding styles applied by sink classes. Additionally, sink styles are commonly not applied in the admin editor due to how WordPress builds markup inside the editor.

#### Updating the front-end to use the `:where()` selector
As discussed [here](https://developer.mozilla.org/en-US/docs/Web/CSS/:where), the `:where()` selector always has 0 specificity. This allows us to easily set very low level styles that can easily be overridden.

**Before:**
```
.t-sink > selector
```
**After:**
```
:where(.t-sink) selector
```

#### Updating the admin preview to use the `:is()` selector
As discussed [here](https://developer.mozilla.org/en-US/docs/Web/CSS/:is), the `:is()` selector includes the specificity of the **_highest regarded argument**. This was required for admin styles which could easily be overridden by WordPress core styles otherwise.

One note here is that I am using the `editor-styles-wrapper` class instead of a sink class. Ideally we may be able to remove sink classes altogether from the admin preview at some point by just targeting classes that WordPress includes out of the box.

```
:is(.editor-styles-wrapper) selector
```

### The Core Columns Block

This PR removes the core Columns block from the deny list. All settings from the Columns block should be supported including up to 6 columns (the max allowed); custom width columns; stacking and not stacking columns on mobile; all alignment settings.

The following blocks should be supported as children. Other blocks have not been tested.

**Supported Child Blocks:**
- Headings
- Paragraphs
- Images
- Lists

### New Core Block Asset Organization
Unlike core blocks such as the Heading block, Paragraph block, and other basic blocks, the Columns block requires a fair amount of styling to support. Because of this, I have created a new [core_columns directory](https://github.com/moderntribe/square-one/tree/74e5158baaa4a2c0589784786f330012bdf96fa5/wp-content/themes/core/components/blocks/core_columns) in the main `theme/components/blocks` repository. This helps better organize CSS and JavaScript specific to this block. The prefixed directory structure helps to identify core blocks as opposed to custom blocks.

## QA

Links to relevant issues
- [Link to Issue](https://moderntribe.atlassian.net/browse/SQONE-609)

- [Post Demo](https://sq1-pr1009.moderntribe.qa/columns-block-posts/?secret=1)
- [Pages Demo](https://sq1-pr1009.moderntribe.qa/columns-block-pages/?secret=1)

Screenshots/video:

## Tests

Does this have tests?

- [ ] Yes


wp-content/themes/core/assets/css/src/theme/media-queries/_variables.pcss
```diff
@@ -10,6 +10,7 @@
 @custom-media --viewport-xxsmall (min-width: 320px);
 @custom-media --viewport-xsmall (min-width: 500px);
 @custom-media --viewport-small (min-width: 600px);
+@custom-media --viewport-medium-max (max-width: 767px);
```
LayaTaal, line 13: Is this worth creating a variable for what is currently a single use case?

wp-content/themes/core/assets/css/src/admin/block-editor/editor.pcss
```diff
@@ -19,18 +19,25 @@
...
 	figure {
 		margin: 0;
 	}
 }
 
+/**
+ * Reset left and right margin for full aligned blocks
+ * to overcome aggressive WordPress core styles.
+ */
+html :where(.editor-styles-wrapper) {
```
LayaTaal, line 32: Unrelated to columns work, but it seems like WordPress changed how they are selecting this so full-align blocks are being left justified as seen [in this screenshot.](http://p.tri.be/fNY1Rj)

wp-content/themes/core/assets/css/src/theme/typography/anchors.pcss
```diff
@@ -32,7 +32,8 @@ a {
  * ------------------------------------------------------------------------- */
 
 .anchor,
-.t-sink a {
+:where(.t-sink) a,
+:is(.editor-styles-wrapper) a {
```
LayaTaal, line 36: Originally, we discussed using `:where(.editor-styles-wrapper)` to handle admin styles. This would be ideal, but essentially has no effect on admin styles because WordPress core styles (usually from `common.css`) overwrite sink styles.

I like using `:is()` because we automatically get most sink styles working in the admin out of the box. The alternative, may be to revert this and just don't touch the admin at all.

wp-content/themes/core/assets/css/src/theme/typography/lists.pcss
```diff
@@ -65,6 +66,7 @@ li {}
 			left: 0;
 			font-size: var(--font-size-body-large);
 			line-height: 2rem; /* list item line-height */
+			color: var(--color-primary);
```
LayaTaal, line 69: I noticed we are applying the primary color for stylized `ol` lists. Should that be the case here as well?

wp-content/themes/core/assets/css/src/admin/block-editor/editor.pcss
```diff
@@ -15,19 +15,20 @@
...
 	* {
-		margin-top: 0;
-		margin-bottom: 0;
-	}
-
-	/* CASE: Alignment: Full */
-	.wp-block[data-align="full"] {
-		margin-left: auto;
-		margin-right: auto;
+		margin: 0;
```
LayaTaal, line 18: WordPress styles the full `margin` property in `common.css`. Changing from updating `margin-top` and `margin-bottom` allowed me to overwrite the core styles more easily and remove some extra css. I was also able to remove the now redundant `figure` styles on line 29.

wp-content/themes/core/components/blocks/core_columns/block-editor.pcss
```diff
@@ -0,0 +1,13 @@
...
+/* -----------------------------------------------------------------------------
+ *
+ * Block Editor
+ * Core Columns Block Styles & Overrides
+ *
+ * Styles applicable to this block in the block editor.
+ *
+ * ----------------------------------------------------------------------------- */
+
+.wp-block[data-type^="core/columns"] {}
```
LayaTaal, line 11: I originally had some admin related styles here. Even though there is nothing left, do we want to keep this file as an example for anyone who needs admin styles?

ChrisMKindred (commented):
🌮 
I appreciate all the work that went into this PR. The more core blocks we can support the better in my opinion. I like the idea of less specificity allowing for easier overwriting when needed. This looks good to me but I'd like to get the feedback from a few of the other FE team members before approving.

wp-content/themes/core/assets/css/src/theme/layouts/default.pcss
```diff
@@ -81,6 +81,16 @@ main {
...
 		}
 	}
+
+	& > .wp-block-columns {
+		width: calc(100% - 2 * var(--grid-gutter-small));
+		margin-left: auto;
+		margin-right: auto;
+
+		@media (--viewport-medium) {
+			width: calc(100% - 2 * var(--grid-gutter));
```
dpellenwood, line 91: Probably should be `--grid-gutter-medium` and then move `grid-gutter` to a viewport-full media query as well.

wp-content/themes/core/assets/css/src/theme/media-queries/_variables.pcss
dpellenwood: I'm fine with this, but I think it should be named `--viewport-small-max`. Imho, `--viewport-medium-max` would be 959px. :)

wp-content/themes/core/assets/css/src/theme/typography/anchors.pcss
dpellenwood: I'm a bit concerned that adding 2 or 3 top-level selectors will bloat our compiled CSS files even more. Would you be willing to compare the css file size between this branch and the default `main` branch, just to make sure it's not a significant jump?

Ideally, I think we'd find a way to automatically replace the `t-sink` selector with the `editor-styles-wrapper` selector on build of the admin css file(s). I'm not concerned with tackling that here though.

wp-content/themes/core/assets/css/src/theme/typography/lists.pcss
dpellenwood: I don't think so, no. That's to provide the colored numbers for the stylized OL, per the [Figma file](https://www.figma.com/file/Wj2g7HVyBQWYIuxgNKMCxf/SQ1-Main-Patterns?node-id=5069%3A292).

wp-content/themes/core/components/blocks/core_columns/block-editor.pcss
dpellenwood: Meh. I say delete this by default. No need to carry around and process an empty css partial on every new project.

wp-content/themes/core/components/blocks/core_columns/css/columns.pcss
```diff
@@ -0,0 +1,109 @@
+/* -----------------------------------------------------------------------------
+ *
+ * Core Block: Columns
```
dpellenwood, line 3: Is this all duped from WP Core? It woukd be good to mention in a comment at the top of the file the original location of these styles.

dpellenwood (changes requested):
This is great Jason! A few small suggestions and changes. And, don't forget a changelog.md entry. :)

wp-content/themes/core/assets/css/src/theme/layouts/default.pcss
LayaTaal: @dpellenwood I got this from line 81 above, where we are also using `-grid--gutter` at the medium size viewport. It looks like that may be a mistake. Do we want to go ahead update both of these?

wp-content/themes/core/assets/css/src/theme/typography/anchors.pcss
LayaTaal: @dpellenwood it looks like we have added about 12kb here (from 310kb to 322kb). If that seems acceptable for this iteration, I can backlog a ticket to improve upon this.

dpellenwood: Thanks for checking. You're good to merge this as-is. We already have some backlog tickets to look into breaking up our monolithic assets into smaller chunks. Those are bigger lifts down the road.

wp-content/themes/core/assets/css/src/theme/layouts/default.pcss
LayaTaal: Updated!

dpellenwood (approved):
🌴 💨

PR #1011, opened 2022-07-01
Field Name and Icon Consistency 

ChrisMKindred:
## What does this do/fix?

Updates custom block field names, field order, and descriptions to be more consistent.
Adds unique custom block icons.
Adds Custom block category.
Removes all instances of the image_select field in favor of the button_group.

You can see the block updates at [sq1-pr1011.moderntribe.qa](http://sq1-pr1011.moderntribe.qa/?secret=1)

**New Block Icons:** 
Screen Shot 2022-07-20 at 6 58 42 PM

## Tests

Does this have tests?

- [ ] Yes
- [X] No, this doesn't need tests I haven't had time to write them yet...


dpellenwood (approved):
:shipit:

wp-content/plugins/core/src/Blocks/Block_Category.php
```diff
@@ -0,0 +1,29 @@
...
+	 * Adds the Custom Category to the front of the block list categories.
+	 *
+	 * @param array $categories
+	 *
+	 * @return array
+	 */
+	public function custom_block_category( array $categories ): array {
+		return array_merge( [
+			[
+				'slug'  => 'tribe-custom',
```
defunctl, line 22: We should make this a constant in this class so it can be used across the other blocks and also updated in a single place so a developer doesn't have to edit every single block after this if we need to change the slug.

wp-content/plugins/core/src/Blocks/Block_Category.php
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Class Block_Category
+ *
+ * Edits the block categories.
+ */
+class Block_Category {
+
+	/**
+	 * Adds the Custom Category to the front of the block list categories.
+	 *
```
defunctl, line 14: Add `/* @filter block_categories */`

defunctl (changes requested):
Missing a changelog entry as well

defunctl:
@ChrisMKindred also don't forget we'll need a companion PR for this in tribe-libs to also add this for new blocks, somewhere in here: https://github.com/moderntribe/tribe-libs/blob/master/src/Generators/templates/block/config.php.tmpl#L28

wp-content/plugins/core/src/Blocks/Block_Category.php
ChrisMKindred: Done.

wp-content/plugins/core/src/Blocks/Block_Category.php
ChrisMKindred: Done.

ChrisMKindred:
Thanks, @defunctl, I have taken care of those items.

[tribe-lib PR for Generator Update](https://github.com/moderntribe/tribe-libs/pull/122)

wp-content/plugins/core/src/Blocks/Blocks_Subscriber.php
```diff
@@ -41,6 +41,10 @@ public function register(): void {
 				$style_override->register();
 			}
 		}, 10, 0 );
+
+		add_filter( 'block_categories', function ( array $categories ): array {
+			return $this->container->get( Block_Category::class )->custom_block_category( $categories );
+		}, 10, 2 );
```
defunctl, line 47:
```suggestion
		}, 10, 1 );
```
This is only passing a single variable into the filter.

ChrisMKindred:
@defunctl When I was looked into the second parameter of the `block_categories` hook I noticed that it was depreciated in 5.8 in favor of the `block_categories_all` hook. I updated the parameter count and updated the hook I was using. I also updated a few of the svg icons after a review from design ([Section Nav and Icon Grid](http://p.tri.be/MVwynt))

tarecord (approved):
🚢 💨

PR #1014, opened 2022-07-05
RFC: Block Middleware

defunctl:
## What does this do/fix?

- Adds a Block Middleware system.
- Adds a global color theme system, that can be applied to blocks or inside block fields.
- Replaces the Tribe ACF Post List Field with a Post Loop Field via Block Middleware.
- Replaces the `Colors` class with a `Swatch_Collection` from square1-field-models.

---
I highly suggest you watch the screencasts before reviewing, pasted from the README:

### Video Guides

- [Block Middleware Overview](https://d.pr/K963cu)

#### Color Theme Middleware

- [How to add the Color Theme Middleware to an existing block](https://d.pr/aIi5HY)
- ~~[How to add the Color Theme Middleware to a repeater](https://d.pr/cUAx8b)~~ - outdated.

### Post Loop Field Middleware

- [Post Loop Field Overview](https://d.pr/Px5xSe)
- ~[How to configure the Post Loop Field Middleware](https://d.pr/FicVgp)~ outdated
- [How to use the Post Loop Field in non-block controllers](https://d.pr/UP7fcL)

---
## TODO
- [x] Remove composer Tribe Libs branch version once https://github.com/moderntribe/tribe-libs/pull/126 is released. 

---

## QA

See screencasts above.

## Tests

Does this have tests?

- [x] Yes
- [ ] No, this doesn't need tests because...
- [ ] No, I need help figuring out how to write the tests.

PR #1017, opened 2022-07-07
Block middleware: global color theme

defunctl:
## What does this do/fix?

- Adds a global color theme field via Block Middleware

## QA

See screencasts in the main [PR](https://github.com/moderntribe/square-one/pull/1014).

## TODO
- [ ] Revert composer.json changes and do an upgrade once https://github.com/moderntribe/square1-field-models/pull/6 is released.

## Tests

Does this have tests?

- [x] Yes


dpellenwood (commented):
This looks good to me with one comment, Justin. We already have a Colors definer class (array) for the whole theme. However, when we updated to this version of Sq1 (using the definers & subscribers), it created a headache of overhead to use these colors b/c each time the class has to be instantiated to access the properties & methods. The previous version of the Colors class was entirely static allowing them to be used with less boilerplate.

I would love it if we could utilize a single static colors class for all things colors/swatches/names throughout the system rather than having colors defined in multiple places.

defunctl:
@dpellenwood 

> This looks good to me with one comment, Justin. We already have a [Colors definer class](https://github.com/moderntribe/square-one/blob/main/wp-content/plugins/core/src/Theme/Config/Colors.php) (array) for the whole plugin + theme. 

Great point and I totally forgot about this class (even though I've wanted to rename it to a proper class name for some time now). 

Is it safe to say that when designers are designing palettes/themes for core blocks/sq1 blocks they're all going to share this set of swatches and there won't be situations where the Color Theme would be different to the palettes defined here?

A tad worried about coupling the middleware to this class, only because it's sharing responsibilities with core WP stuff and as of right now, the Color Theme Middleware's functionality is mostly sitting in its folder, making it a bit more portable (if you have the core middleware system implemented).

My worry is basically a developer changes the swatches in the Theme_Definer and that causes unexpected regressions. If we truly think that is coupled together with this color theme middleware though, I think we can make this work.

The Colors class is a pretty well defined class though, so perhaps moving the class itself into Tribe Libs that instead implements an interface in order to:

1. Allow it to be shared across the code base, knowing it will be there for both Theme Support and Middleware.
2. If we make it implement an interface, and pass that interface to controllers instead of the class, this allows developers in their forked projects to create their own class to override this and one line in PHP-DI to tell the entire code base to use their custom Color class anywhere PHP-DI expects it.

I'll play with this a bit.

> However, when we updated to this version of Sq1 (using the definers & subscribers), it created a headache of overhead to use these colors b/c each time the class has to be instantiated to access the properties & methods. The previous version of the Colors class was entirely static allowing them to be used with less boilerplate.
> I would love it if we could utilize a single static colors class for all things colors/swatches/names throughout the system rather than having colors defined in multiple places.

Utilizing a static instance all through the code base is definitely a code smell.

We shouldn't ever need to manually instantiate this, that's PHP-DI's job. Simply injecting it into the constructor of a class and assigning it to a property is all the work that would need to be done.

That way we're never polluting the global scope, as static calls to methods across the code base is basically "globals in disguise" and makes all those classes using them untestable, because we can't mock that static instance which we could easily do if it's just injected into the constructor. 

In addition to this, if we go with the suggestion of making it implement an interface above, developers can switch out the entire instance without touching the classes using them with a single line in a definer, giving them the flexibility to override the underlying functionality should any project scope change drastically. 

If this was a static class called across many controllers/methods, refactoring becomes a really dangerous exercise, forcing the developer to potentially edit all the classes calling the static instance.

Hope this makes sense/is a good explanation!

PR #1019, opened 2022-07-13
Fix: Unsupported Browser Favicon showing the Tribe logo

verytwisty:
## What does this do/fix?

On the Unsupported browser page, remove tribe favicon and make the favicon the site icon chosen in the customizer

## QA

QA ticket from Military Affairs:
https://moderntribe.atlassian.net/browse/MSSA-268


wp-content/themes/core/routes/unsupported_browser/Unsupported_Browser_Controller.php
```diff
@@ -29,6 +29,14 @@ public function get_legacy_image_url( string $filename ): string {
 		return esc_url( trailingslashit( get_template_directory_uri() ) . 'assets/img/theme/legacy-browser/' . $filename );
 	}
 
+	public function get_favicon_url(): string {
+		if ( get_site_icon_url() ) {
+			return esc_url( get_site_icon_url() );
```
defunctl, line 34: This function returns a string but it can be filtered so you can just replace this whole method with `return (string) get_site_icon_url();` The `if` conditional isn't needed

wp-content/themes/core/routes/unsupported_browser/unsupported_browser.php
```diff
@@ -18,7 +18,10 @@
 
 	<?php echo $c->get_styles(); ?>
 
-	<link rel="shortcut icon" href="<?php echo esc_url( trailingslashit( get_template_directory_uri() ) . 'assets/img/theme/branding-assets/favicon.ico' ); ?>">
+	<?php if ( $c->get_favicon_url() ) { ?>
+		<link rel="shortcut icon" href="<?php echo $c->get_favicon_url(); ?>">
```
defunctl, line 22:
```suggestion
		<link rel="shortcut icon" href="<?php echo esc_url( $c->get_favicon_url() ); ?>">
```
We should escape as late as possible, so in the view when we can :smile:

defunctl (changes requested):
We also need a changelog entry if you want to add something to the top of CHANGELOG.md :)

wp-content/themes/core/routes/unsupported_browser/Unsupported_Browser_Controller.php
dpellenwood: If we're going to ue WP's icons, we should probably just render `wp_site_icon()` in the template directly. IIRC, using `get_site_icon_url()` will only a single URL for the original site icon which probably isn't compatible with old browsers. We want to make sure we're serving at least the 32px .png file for old browsers.

dpellenwood (approved):
Thanks for grabbing this, @verytwisty . I made one suggestion, and Justin's comments are good.

Fwiw, I'm going to backlog a ticket to remove the "Unsupported Browser" feature from SquareOne entirely in the near future. It's only "blocking" visitors using IE9 or lower, which is basically no one in the world any more.

ChrisMKindred (commented):
@dpellenwood is this PR still relevant now that we are completely removing the Unsupported Browser Functionality (#1023 )?

dpellenwood:
> @dpellenwood is this PR still relevant now that we are completely removing the Unsupported Browser Functionality (#1023 )?

It is not. We can close this PR with a hat-tip to @verytwisty for raising this issue which led to removing the feature entirely. 🙇

PR #1020, opened 2022-07-19
RCF: General Settings Updates

dpellenwood:
## What does this do/fix?

This is an alternative approach to PR #1012 where I proposed using the Customizer for all "theme settings". In ths PR, I'm proposing that we don't use the Customizer and instead continue using ACF fields for the settings. However, I'm suggesting that we should:
* Move our Settings > General Settings page to Appearance > [Theme Name] Options (with a fixed slug).
* Migrate Analytics and Social Media settings to Tabs within the Theme Options to provide room for additional theme options tabs.

This is an interim change to how we handle theme settings/options. While the Customizer is a more "WordPress-y" location for these types of fields, the Customizer's field API is very restrictive and will slowly be eliminated in the era of Full Site Editing. Long-term we hope to embrace FSE for many of these fields.

## Screenshots
* [Analytics Setting Tab](http://p.tri.be/nVeYMw)
* [Social Settings Tab](http://p.tri.be/SEcpo8)

## Tests

Does this have tests?

- [ ] Yes


wp-content/plugins/core/src/Object_Meta/Theme_Options.php
```diff
@@ -0,0 +1,100 @@
...
+			'type'  => 'tab',
+		] );
+
+		return $tab_field;
+	}
+
+	private function get_analytics_gtm_id_field(): ACF\Field {
+		$field = new ACF\Field( self::NAME . '_' . self::ANALYTICS_GTM_ID );
+		$field->set_attributes( [
+			'label'       => __( 'Google Tag Manager ID', 'tribe' ),
```
defunctl, line 80:
```suggestion
			'label'       => esc_html__( 'Google Tag Manager ID', 'tribe' ),
```

wp-content/plugins/core/src/Settings/Theme_Options.php
```diff
@@ -0,0 +1,30 @@
...
+
+use Tribe\Libs\ACF\ACF_Settings;
+
+class Theme_Options extends ACF_Settings {
+
+	public const SLUG = 'theme-options';
+
+	public function get_title(): string {
+		$current_theme = wp_get_theme();
+		$theme_name    = $current_theme->exists() ? $current_theme->get( 'Name' ) : 'Theme';
```
defunctl, line 13: I wonder if we should just simplify this to not use the theme name and just standardize is across all projects to `Settings` or `Theme Options` or `Theme Settings`. 

That way, if we need to write documentation, help developers or clients find this page, it's always named the same thing.

This also points out a grammar error in our stub theme name, should be `Your Client's Name` :smile:

wp-content/themes/core/components/follow/Follow_Controller.php
```diff
@@ -51,22 +50,22 @@ public function get_social_links(): Social_Link_Collection {
 
 	protected function get_label( string $site ): string {
 		switch ( $site ) {
-			case Social_Settings::FACEBOOK:
+			case Theme_Options::SOCIAL_FACEBOOK:
 				return __( 'Like us on Facebook', 'tribe' );
```
defunctl, line 54:
```suggestion
				return esc_html__( 'Like us on Facebook', 'tribe' );
```

Might as well fix all of these while we're in here, the ones below also need the same treatment.

wp-content/plugins/core/src/Object_Meta/Theme_Options.php
```diff
@@ -0,0 +1,100 @@
...
+	 *
+	 * @return mixed|null
+	 */
+	public function get_value( $key, $post_id = 'option' ) {
+		return in_array( $key, $this->get_keys(), true ) ? get_field( $key, $post_id ) : null;
+	}
+
+	public function get_group_config(): array {
+		$group = new ACF\Group( self::NAME, $this->object_types );
+		$group->set( 'title', __( 'Theme Options', 'tribe' ) );
```
defunctl, line 48:
```suggestion
		$group->set( 'title', esc_html__( 'Theme Options', 'tribe' ) );
```

The remaining labels in here also need escaping.

defunctl (changes requested):
Loving the tabs, nice work.

wp-content/plugins/core/src/Settings/Theme_Options.php
dpellenwood: Yeah, Chris & I discussed that a bit. I'd prefer to use the actual theme name for the label to make it clear it's not a generic page and instead specific to the project/client. The same issue we have with "General Settings" would apply to "Theme Options" imho.

ChrisMKindred (approved):
👍

wp-content/plugins/core/src/Object_Meta/Theme_Options.php
```diff
@@ -0,0 +1,100 @@
...
+		return $tab_field;
+	}
+
+	private function get_analytics_gtm_id_field(): ACF\Field {
+		$field = new ACF\Field( self::NAME . '_' . self::ANALYTICS_GTM_ID );
+		$field->set_attributes( [
+			'label'       => esc_html__( 'Google Tag Manager ID', 'tribe' ),
+			'name'        => self::ANALYTICS_GTM_ID,
+			'type'        => 'text',
+			'placeholder' => esc_html__( 'Enter Google Tag Manager ID (GTM-XXXX)', 'tribe' ),
```
defunctl, line 83:
```suggestion
			'placeholder' => esc_attr__( 'Enter Google Tag Manager ID (GTM-XXXX)', 'tribe' ),
```
Whoops, this is an attribute.

defunctl (approved):
Just one escaping change but this looks good to me.

PR #1027, opened 2022-07-28
[SQONE-708]: Fix for TinyMCE floating toolbar repositioning infinite loop.

bjcooper:
## What does this do/fix?

Fixes an issue where the TinyMCE floating/inline toolbar (usually when editing a link in the WYSIWYG) "flickers" endlessly. This is especially likely to happen when the TinyMCE editor is inside of a repeater (as in the Accordion block, for example) due the restricted horizontal space.

The issue is ticketed here: https://core.trac.wordpress.org/ticket/44911 and everyone pretty much agrees that it exists, but no one has fixed it in ~3 years.

The issue is caused by some poor positioning logic in `wp-includes/js/tinymce/plugins/wordpress/plugin.js`, which sometimes places the floating toolbar slightly off-screen when positioning it next to the inline element being edited. Overflowing the body this way causes horizontal scrollbars to appear and triggers a `scrollwindow` event. The same plugin hides the toolbar during scroll events to be repositioned over the inline element being edited once the scrolling finishes. Hiding the toolbar causes the scrollbars to disappear and the plugin then shows the toolbar again, triggering the scrollbars, triggering the hiding of the toolbar, etc, etc.

The fix is rather simple, but difficult to apply. It would essentially be a one-liner in the core TinyMCE Wordpress plugin, were that a normal library you could patch or replace. 😢 Instead, I've created a custom TinyMCE plugin to observe the toolbar and fix the positioning in the trouble cases. TinyMCE plugins are meant to be independent and "closed", so there's no way to fix the errant positioning function directly. Instead we quickly recalculate and reposition the element ourselves after the bad positioning has been done.

## QA

The ticket for this PR: https://github.com/moderntribe/turner-industries/pull/78
Original fix forMilitary Affairs: https://github.com/Microsoft-Stories/military-affairs/pull/221
Subsequent fix for Turner Industries: https://github.com/moderntribe/turner-industries/pull/78

Screenshots/video:
image

Important note:
This can potentially affect any inline WYSIWYG toolbar, so the potential effects of some bad code here are substantial. Also, there's already some existing "weirdness" with the TinyMCE editor toolbar positioning in some scenarios that isn't related to this fix—this fix should just make the toolbar (at-least) usable in all cases.

## Tests

Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests because I don't know how you'd test it. Maybe it does?

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -3,6 +3,7 @@
 All notable changes to this project will be documented in this file.
 
 ## 2022.07
+* Fixed: TinyMCE floating toolbar repositioning loop.
 * Added: Block Category (Custom) for all custom blocks.
 * Updated: Block fields layout and language for consistency.
 * Updated: Removed `image_select` fields in favor of `button_group` fields on blocks.
--- a/wp-content/themes/core/assets/js/src/admin/editor/tribe-tinymce.js
+++ b/wp-content/themes/core/assets/js/src/admin/editor/tribe-tinymce.js
@@ -4,9 +4,9 @@
  * This is especially likely to happen when the TinyMCE editor is inside of a
  * repeater (as in the Accordion block, for example) due the restricted
  * horizontal space.
- * 
+ *
  * The issue is ticketed here: https://core.trac.wordpress.org/ticket/44911.
- * 
+ *
  * The issue is caused by some poor positioning logic in
  * `wp-includes/js/tinymce/plugins/wordpress/plugin.js`, which sometimes places
  * the floating toolbar slightly off-screen when positioning it next to the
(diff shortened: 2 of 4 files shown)


changelog.md
```diff
@@ -3,6 +3,7 @@
 All notable changes to this project will be documented in this file.
 
 ## 2022.07
+* Fixed: TinyMCE floating toolbar repositioning loop.
```
dpellenwood, line 6: Would you kindly add a link to the related WP Core issue here, for reference?

wp-content/themes/core/assets/js/src/admin/editor/tribe-tinymce.js
```diff
@@ -0,0 +1,64 @@
+( function() {
```
dpellenwood, line 1: Kindly add a DocBlock header explaining the purpose and linking to the related WP Core issue? Future-us will thank you for the references.

dpellenwood (approved):
Thanks, Brett! This is great. Just a couple documentation asks, please.

PR #1029, opened 2022-08-03
Bugfix: PHPStan have_posts() false positive

defunctl:
## What does this do/fix?

- Properly tells PHPStan that the `have_posts()` function is impure

## QA

The following will no longer be tagged as `While loop condition is always true.` or `Unreachable statement - code above always terminates.`:

```php
			<?php
			if ( have_posts() ) : ?>
				<div class="archive-loop__items g-3-up">
					<?php while ( have_posts() ) :
						the_post();
						get_template_part( 'components/card/card', '', $c->get_extended_card_args(
							$c->get_meta_primary(),
							$c->get_meta_secondary()
						) );
					endwhile; ?>
				</div>
				<?php get_template_part( 'components/pagination/loop/loop' );
			else :
				get_template_part( 'components/no_results/no_results' );
			endif;
			?>
```

## Tests

Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests because it's a change to code analysis configuration.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,9 @@
 
 All notable changes to this project will be documented in this file.
 
+## 2202.08
+* Fixed: False positive on `have_posts()` using PHPStan if proceeded by an `if ( have_posts() ):`, utilizing a new [wordpress-overrides.stub file](dev/stubs/wordpress-overrides.stub)  
+
 ## 2022.07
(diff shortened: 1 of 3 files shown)


ChrisMKindred (approved):
:shipit:

PR #1030, opened 2022-08-06
POC: Block Pattern

ChrisMKindred:
## What does this do/fix?

An example of simplifying one of our custom blocks and using core blocks in a block pattern to handle the layout. In this example, I have recreated a content loop block by creating an automatic loop block using ACF and WordPress Core blocks to fill out the other fields.  

You can play with how block patterns are handled in the UI here: [https://sq1-pr1030.moderntribe.qa/pattern-test-post/](https://sq1-pr1030.moderntribe.qa/pattern-test-post/?secret=1)

## Tests

Does this have tests?

- [ ] Yes
- [X]  No, this doesn't need tests because it is a proof of concept.


wp-content/plugins/core/src/Blocks/Patterns/Loop_Pattern.php
```diff
@@ -0,0 +1,46 @@
...
+			self::TITLE       => esc_html__( 'Loop Pattern', 'tribe' ),
+			self::DESCRIPTION => esc_html__( 'Loop Pattern', 'tribe' ),
+			self::CONTENT     => $this->get_content(),
+			self::CATEGORIES  => [ Pattern_Category::CUSTOM_PATTERN_CATEGORY_SLUG ],
+			self::KEYWORDS    => [ esc_html__( 'loop', 'tribe' ), esc_html__( 'query', 'tribe' ) ],
+		];
+	}
+
+	private function get_content(): string {
+		return '<!-- wp:group {"tagName":"section","className":"c-block b-content-loop l-container"} -->
```
LayaTaal, line 25: @ChrisMKindred would we be able to create a template file for this and return it here using something like `tribe_template_part`? It would be great to make this easier to read and edit.

ChrisMKindred: @LayaTaal 
That is an option for sure. I don't know if, in practice, you would ever be interacting with the content directly in your code editor. The way I see it is that Design and Engineering would work together on setting up the blocks on a page. Once everybody is happy with the layout, switch to the Code Editor in the block editor settings and copy the content and paste it into whatever we are using to return the pattern content.

PR #1033, opened 2022-08-10
Remove Configure SMTP plugin

tarecord:
## What does this do/fix?

On a current project Configure SMTP throws errors with every request and ends up triggering the Whoops UI, making local development a pain. This swaps that plugin (which was [last updated 11 years ago](https://wordpress.org/plugins/configure-smtp/)) with [a plugin that is actively maintained](https://wordpress.org/plugins/wp-mail-smtp/) and offers a lot of functionality.


ChrisMKindred:
@tarecord I like that configure SMTP doesn't try to throw a pro version in your face or take up space in the main WP Nav.  What is the error that is getting thrown?

composer.json
```diff
@@ -117,6 +116,7 @@
     "wpackagist-plugin/regenerate-thumbnails": "3.1.5",
     "wpackagist-plugin/user-switching": "1.6.0",
     "wpackagist-plugin/wordpress-seo": "19.3",
+    "wpackagist-plugin/wp-mail-smtp": "^3.5",
```
defunctl, line 119: @tarecord I know you didn't ask for my review, but you should add the specific version for plugins so during deploys a new version doesn't just get randomly installed without being tested.

dpellenwood:
Just a thought...do we _need_ to require a SMTP plugin by-default in SquareOne?

ChrisMKindred:
> @dpellenwood  Just a thought...do we need to require a SMTP plugin by default in SquareOne?
No

defunctl:
> Just a thought...do we _need_ to require a SMTP plugin by default in SquareOne?

Probably not, nope.

tarecord:
> Just a thought...do we need to require a SMTP plugin by default in SquareOne?

That's what I'm thinking. I figured having a concrete PR to bubble up discussion would be good :)

@ChrisMKindred I'm seeing this when its activated:
```
Declaration of c2c_ConfigureSMTP::options_page_description() should be compatible with C2C_Plugin_023::options_page_description($localized_heading_text = '')
```

@dpellenwood if I pull out the default, do we have a list of recommended SMTP plugins?

dpellenwood:
> @dpellenwood if I pull out the default, do we have a list of recommended SMTP plugins?

I generally point client's to their hosting provider's recommendation for email/SMTP. Here is [Pantheon's](https://pantheon.io/docs/email) and [WP Engine's](https://wpengine.com/support/using-3rd-party-email-provider-send-mail-wordpress/). In short, (in order of recommendation), I would say:
1. [MailGun](https://wpengine.com/support/configuring-mailgun-for-wordpress/) another providers which supports API-based connections
2. [WP Mail SMTP](https://wordpress.org/plugins/wp-mail-smtp/) for a true SMTP connection. (Recommended by both Pantheon & WPE.)

ChrisMKindred (changes requested):
Let's just remove the plugin. Each site has its own unique email needs (MailGun, SMTP, Hosting requirements, etc...), and the decision should be based on that. We also have documentation that provides a list of plugins we recommend using for those situations.

dpellenwood:
> Let's just remove the plugin. Each site has its own unique email needs (MailGun, SMTP, Hosting requirements, etc...), and the decision should be based on that. We also have documentation that provides a list of plugins we recommend using for those situations.

@tarecord , curious if you'd have a chance to get back to this update the PR to remove this plugin from our default composer?

tarecord:
> @tarecord, curious if you'd have a chance to get back to this update the PR to remove this plugin from our default composer?

@dave, this should be good to go now.

dpellenwood (approved):
Looks good Tanner, just a changelog entry then you can merge. Thanks!

ChrisMKindred (approved):
👍 Looks good, thanks!

PR #1037, opened 2022-08-18
[SQONE-698]: Basic PHP80 support

defunctl:
## What does this do/fix?

> This is an intermediate update to show how **existing PHP7.4 projects** can jump to PHP8.0 when their hosts force an upgrade. 
> Essentially, this is the minimum amount of work to make this support PHP8.0 while being backwards compatible with our PHP7.4 system.
> A future PR will bring in PHP8.0 minimums + optimizations once Tribe Libs is also ready for that.

* Changed: Added basic support for PHP8.0. 
* Removed: The Tribe Admin Dashboard plugin (not PHP8.0 compatible).
* Bumped: Tribe Libs to the `4.x` major version + plugin updates.
* Removed: Any dependencies that Tribe Libs already provides in `composer.json`.
* Fixed: PHPCS throwing `trim()` errors on certain files due to bad WordPress rule typing.
* Changed: Replaced [Function Mocker](https://github.com/lucatume/function-mocker) with [Brain Monkey](https://brain-wp.github.io/BrainMonkey/) for Unit testing as Function Mocker isn't fully PHP8.0 compatible.
* Added: New [Unit Test Class](dev/tests/tests/_support/Classes/Unit.php) that configures Brain Monkey.
* Updated: Existing Unit tests to use Brain Monkey where they were using Function Mocker.

## QA

Everything should function as before, assuming you restart the project to run on the PHP8.0 docker container.


phpcs.xml.dist
```diff
@@ -1,32 +1,35 @@
...
-    <!-- Exclude files -->
-    <exclude-pattern>*-config.php</exclude-pattern>
-    <exclude-pattern>*vendor/</exclude-pattern>
-    <exclude-pattern>*tests/*</exclude-pattern>
-    <exclude-pattern>*.twig</exclude-pattern>
-    <exclude-pattern>*webpack/</exclude-pattern>
+	<!-- Directories to be checked -->
+	<file>./wp-content/plugins/core</file>
+	<file>./wp-content/themes/core</file>
+	<file>./wp-content/mu-plugins</file>
```
ChrisMKindred, line 24: 💡 
Since some hosts require mu-plugins, would it be worthwhile for us to be more specific here?

ChrisMKindred (approved):
A great example for migrating!  Thanks @defunctl

phpcs.xml.dist
defunctl: As we discussed on slack, it would be better for this to fail when bringing in a 3rd party plugin and having the developer have to add it to the ignore list there than it not silently scanning a custom mu-plugin we're writing.

ChrisMKindred: 👍

PR #1041, opened 2022-08-28
Fix: Share Linkedin URL

alexander-kozhukhovski:
## What does this do/fix?

- Replaces `__()` with `esc_html__()` in `wp-content/themes/core/components/share/Share_Controller.php`
- Replaces Linkedin share URL with a new one

## Tests

Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests

### DIFF
--- a/wp-content/themes/core/components/share/Share_Controller.php
+++ b/wp-content/themes/core/components/share/Share_Controller.php
@@ -113,7 +113,8 @@ class Share_Controller extends Abstract_Controller {
 		} elseif ( is_search() ) {
 			$query = get_search_query();
 
-			$data['link']  = $this->normalize_url( get_search_link( $query ) );
+			$data['link'] = $this->normalize_url( get_search_link( $query ) );
+			/* translators: %s: Search results. */
 			$data['title'] = sprintf( esc_html__( 'Search Results: %s', 'tribe' ), esc_attr( $query ) );
 		}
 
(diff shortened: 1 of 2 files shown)


wp-content/themes/core/components/share/Share_Controller.php
```diff
@@ -114,7 +114,7 @@ private function get_data(): array {
 			$query = get_search_query();
 
 			$data['link']  = $this->normalize_url( get_search_link( $query ) );
-			$data['title'] = sprintf( __( 'Search Results: %s', 'tribe' ), esc_attr( $query ) );
+			$data['title'] = sprintf( esc_html__( 'Search Results: %s', 'tribe' ), esc_attr( $query ) );
```
ChrisMKindred, line 117: it is always good to include a comment directed at translators when using `sprint()`.  That way the translator doesn't have to know code to figure out what they are translating. 

Here is an [example from the WP handbook](https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/#variables).

ChrisMKindred (approved):
Thanks!  I have a quick comment to check out but if you don't have additional time, feel free to merge.

dpellenwood (approved):
Looks good. Thanks, Alex!

wp-content/themes/core/components/share/Share_Controller.php
alexander-kozhukhovski: Yes, good point. I updated with a comment.

PR #1042, opened 2022-08-30
Bugfix/post loop middleware unique field names

defunctl:
## What does this do/fix?

- ACF's field store will store a field by key, so each key needs to be unique, even if under a group. When putting two blocks with Post Loop Field Middleware on the same page, the configuration was being overwritten/shared due to non-unique key names.
- Post Loop Middleware keys are now prefixed with the block name.
- Fixed a bug in the Content Loop Block where there was no check if the post has a category, because one can select pages and they don't have categories.

## QA

- Post Loop configuration should now be properly isolated to each block

## Tests

Does this have tests?

- [x] Yes

### DIFF
--- a/wp-content/themes/core/components/blocks/content_loop/Content_Loop_Controller.php
+++ b/wp-content/themes/core/components/blocks/content_loop/Content_Loop_Controller.php
@@ -116,7 +116,7 @@ class Content_Loop_Controller extends Abstract_Controller {
 			$cards[] = [
 				Card_Controller::STYLE           => $layout,
 				Card_Controller::USE_TARGET_LINK => (bool) $post->cta->link->url,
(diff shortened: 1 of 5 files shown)


wp-content/themes/core/components/blocks/content_loop/Content_Loop_Controller.php
```diff
@@ -245,7 +245,11 @@ private function get_content(): Deferred_Component {
 		] );
 	}
 
-	private function get_card_meta_primary( string $cat_name ): Deferred_Component {
+	private function get_card_meta_primary( string $cat_name ): ?Deferred_Component {
```
alexander-kozhukhovski, line 248: @defunctl Thank you

alexander-kozhukhovski (approved):
Looks good for me. Thank you! 🌮

PR #1047, opened 2022-09-21
Bugfix: Follow component social icon CSS class

defunctl:
## What does this do/fix?

- Using the Follow component would not output any social icons because the CSS class names were incorrect.

## QA

The icons now show up when using the Follow component:

image

## Tests

Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests because, well it probably could have used them.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -3,6 +3,7 @@
 All notable changes to this project will be documented in this file.
 
 ## 2022.09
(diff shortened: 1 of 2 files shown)


dpellenwood (approved):
Ah dang. I caused this whrn I update the theme settings. Good catch, Justin.

PR #1048, opened 2022-09-23
Bugfix: Single Controller `get_author_id()` now returns an integer type instead of a string.

defunctl:
## What does this do/fix?

- Although WP Query says the `post_author` property is a string for compatibility reasons, functions using it expect it to be an integer.

I'm not sure why PHPSTAN allowed this through in the first place?

## QA

image


ChrisMKindred (approved):
Thanks @defunctl 👍

PR #1053, opened 2022-10-18
Bugfix: Term Sorter

lindseyolson:
## What does this do/fix?
- Adds a fix in the Term Sorter Plugin for the `wp_localize_script()` method: 3rd argument is expected to be an array, not a string
- Error is only seen locally when debug is on

## Tests

Does this have tests?

- [ ] Yes
- [X] No, this doesn't need tests because it's a bugfix


dpellenwood (approved):
Good catch, Lindsey!

dpellenwood:
@lindseyolson , don't forget to add a changelog entry, too please.

ChrisMKindred (approved):

PR #1055, opened 2022-10-18
Update/misc js updates

dpellenwood:
## What does this do/fix?
Updates to our body lock/unlock utility script to address the new(ish) smooth scroll behavior available in browsers and to point to an MT-owned (and updated) copy of the accessible dialog library. 

Note that the changes to the body lock/unlock script only affect sites that set `scroll-behavior: smooth` on the html or body elements to enable smooth scrolling for the whole document.

The a11y-dialog update is something that should be addressed on any project using the script as the previous repo/version has a bug related to the scrollTop not being calculated correctly in newer versions of Safari. 

## QA

These issues were raised via a ticket in the Harvard project.
- [HMNT22-58](https://moderntribe.atlassian.net/browse/HMNT22-58)

## Tests

Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests because this is a JS update.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -3,6 +3,7 @@
 All notable changes to this project will be documented in this file.
 
 ## 2022.10
+* Updated: Switched to an mT-owned version of a11y-dialog w/ misc bug fixes and update body lock/unlock scripts to address CSS-defined `scroll-behavior: scroll`. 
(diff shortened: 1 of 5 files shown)


jamiepaul (approved):
👍 👍

ChrisMKindred (approved):
👍

LayaTaal (approved):
Thanks Dave!

PR #1058, opened 2022-12-09
Add .deploy-ignore file and update deploy workflows

defunctl:
## What does this do/fix?

- Add a single configuration file for all files/folders to ignore when using rsync to sync a built project's wp-content folder.

## QA

Any of these files inside of wp-content should never get deployed to a server.

## Tests

Does this have tests?

- [ ] Yes
- [x] No, this doesn't need tests because it's GitHub workflow update.

### DIFF
--- a/.deploy-ignore
+++ b/.deploy-ignore
@@ -2,8 +2,8 @@
 # This file is run via GitHub workflows, see the .github/workflows folder.
 
 # This file and any include files.
-.rsyncignore
-.rsyncinclude
+.deploy-ignore
+.deploy-include
 
 # Large/disallowed file types on WP Engine
 *.hqx
--- a/.github/workflows/deploy-dev.yml
+++ b/.github/workflows/deploy-dev.yml
@@ -108,7 +108,7 @@ jobs:
       # Rsync plugin and theme files
       - name: Sync Build WP-CONTENT folder to Deploy directory
         run: |
-          rsync -rpv --delete --exclude-from="${{env.BUILD_FOLDER}}/.rsyncignore" ${{env.BUILD_FOLDER}}/wp-content ${{env.DEPLOY_FOLDER}} \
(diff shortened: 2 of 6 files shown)


ChrisMKindred (commented):
This is great. There is a common pattern being used by tools like [dist-archive](https://developer.wordpress.org/cli/commands/dist-archive/) and on [VIP](https://docs.wpvip.com/technical-references/development-workflow/automated-build-and-deploy/) calling this file `.deployignore`.  Do you think that makes more sense in this case?

defunctl:
> This is great. There is a common pattern being used by tools like [dist-archive](https://developer.wordpress.org/cli/commands/dist-archive/) and on [VIP](https://docs.wpvip.com/technical-references/development-workflow/automated-build-and-deploy/) calling this file `.deployignore`. Do you think that makes more sense in this case?

Sorry for the delay here!

A few issues with this I can see with renaming this to`.deployignore`:

My only concern with renaming this to`.deployignore` is that file name is primarily associated with [DeployHQ](https://www.deployhq.com/blog/excluding-files-with-deployignore) and even PhpStorm has the automatic file association, which may confuse people editing it.

image

Due to this, PhpStorm has some odd inspections associated with the file type and anything it suggests is likely not compatible with [rsync's pattern format](https://man7.org/linux/man-pages/man1/rsync.1.html#INCLUDE/EXCLUDE_PATTERN_RULES).

image

Note that VIP's deploy script [actually uses git to deploy](https://github.com/Automattic/vip-go-build/blob/master/deploy.sh#L150-L151).

So I guess a compromise, how about `.deploy-ignore` and I'll include a link to the rsync patterns in its comments as well?

ChrisMKindred:
VIP uses a git push. The [issue is that they will not deploy anything](https://docs.wpvip.com/technical-references/development-workflow/automated-build-and-deploy/#h-deploying-built-files-from-gitignore) in the `.gitignore` file.  The `.distignore` allows you to include a built file or remove a src folder from the build. 

`.deploy-ignore` is fine with me. A comment in the file with a link to the rysnc patterns is super helpful.  Good Idea!

ChrisMKindred (approved):

PR #1061, opened 2023-01-20
This check should be looking for an empty value.

Vinsanity:
## What does this do/fix?

Found this one out the hard way. It should be looking for an empty string. Surprised that [this is 5 months old](https://github.com/moderntribe/square-one/blame/ef0d180925a5dfef6d205b33c79157bcbf5330e0/wp-content/themes/core/components/blocks/content_loop/Content_Loop_Controller.php#L151).

Let's get this one merged before someone else get hit with it.


dpellenwood (approved):
🦆 💨

ChrisMKindred (approved):
👍

PR #1063, opened 2023-02-08
Fix for multiple TinyMCE editors on the same page

bjcooper:
## What does this do/fix?

This patch fixes an edge case where there are multiple TinyMCE editors active on a given page (imagine _Appearance → Theme Configuration_ with multiple tabs, several of which have a WYSIWYG field, for example). Without this patch, the `tribe-tinymce` plugin (which itself fixes a different core Wordpress + TinyMCE bug) fails to find the correct editor and breaks TinyMCE floating toolbar positioning (they end up positioned offscreen and unusable).

## QA

Links to relevant issues
- https://moderntribe.atlassian.net/browse/SQONE-719

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -3,6 +3,7 @@
 All notable changes to this project will be documented in this file.
 
 ## 2023.02
+* Fixed: Fixed the TinyMCE floating toolbar fix, which would break with multiple editors active on the same page.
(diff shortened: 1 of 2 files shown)


dpellenwood (approved):
🧊 💨

PR #1065, opened 2023-02-16
Fix php version typo

defunctl:
## What does this do/fix?

- `ci.yml` wasn't actually using the correct PHP version due to a typo.


ChrisMKindred (approved):
👍 Sorry I missed it in the first review!

PR #1067, opened 2023-03-06
Improve remove_editor_style_reset

szepeviktor:
## What does this do/fix?

Improve `remove_editor_style_reset` and aid static analysis the same time


szepeviktor:
I am sorry that this change has less than minimum value.

PR #1070, opened 2023-03-07
Bugfix/builds fail due to browserlist key change

dpellenwood:
## What does this do/fix?
Reverts the browserlist key usage "not dead" in favor of named browsers as the new key doesn't appear supported with our current set of npm packages.


defunctl (approved):
All workflows passing now, thanks @dpellenwood !

PR #1071, opened 2023-04-17
Allow hiding certain embed variation instead of hiding whole embed block

MlKilderkin:
## What does this do/fix?

Allow hiding embed variation of block instead of completely disable embed block
The idea came from HRI project where we should show embed block and Youtube embeds only

## Tests

Does this have tests?

- [ ] Yes

### DIFF
--- a/wp-content/themes/core/assets/js/src/admin/editor/deny-blocks.js
+++ b/wp-content/themes/core/assets/js/src/admin/editor/deny-blocks.js
@@ -1,4 +1,4 @@
-import { getBlockTypes, unregisterBlockType } from '@wordpress/blocks';
+import { getBlockTypes, unregisterBlockType, unregisterBlockVariation } from '@wordpress/blocks';
 import { BLOCK_DENYLIST } from '../config/wp-settings';
 
 /**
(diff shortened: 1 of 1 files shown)


dpellenwood (commented):
Thanks for looking into this, @MlKilderkin . As I understand the Embed block, this just "hides" the embed inserter button in the admin interface. It doesn't actually prevent an editor from dropping an embed URL for any of the oEmbed services into the editor and seeing it transformed into an Embed. Effectively, there's no way to actually eliminate the possibility of embedding from any oEmbed source as long as any subset of variations for embed block are enabled.

Knowing this, I'd prefer to not limit embeds at all rather than selectively "hiding" some sources form the inserter.

MlKilderkin:
> Thanks for looking into this, @MlKilderkin . As I understand the Embed block, this just "hides" the embed inserter button in the admin interface. It doesn't actually prevent an editor from dropping an embed URL for any of the oEmbed services into the editor and seeing it transformed into an Embed. Effectively, there's no way to actually eliminate the possibility of embedding from any oEmbed source as long as any subset of variations for embed block are enabled.
> 
> Knowing this, I'd prefer to not limit embeds at all rather than selectively "hiding" some sources form the inserter.

That's correct. It is just button removal from blocks list. You can still use any embeds. 

No problem with that. I'll just close the PR

PR #1076, opened 2024-09-03
Task/composer required package installs update

dpellenwood:
## What does this do/fix?

* Removes Gravity Forms as a composer dependency as it's doesn't technically work. 
* Moves ACF Pro to use [their recommended approach](https://www.advancedcustomfields.com/resources/installing-acf-pro-with-composer/).


JMRhodes (approved):
👍

